PR #2, opened 2018-06-21
Configure Travis to deploy to the Development Environment

LucasCioffi:


.travis.yml
```diff
@@ -0,0 +1,7 @@
+language: node_js
+node_js:
+- '6'
+script: npm test
+env:
+  global:
+    secure: d7Tpzk8eq+LU/amOUlg8E2QtFalDjDmBeFUzUwJi7qeqKt0EaIKmvJdzYaIJ2B9aYfONuV/mxBLkWVVuQkycI/w88cSRh+7dGOrKKd9xL6iJkVSROt1KAGhlcrXoeJrEwYrSUfgRCz+7tfym321lauXw+4NVn7GKSgvZ4kHyiF9wUQpu/MFZVNzrzLmkdcKAfQVC9du3d/Q5g99rBwhBPme2YsdnhKfiGYFiu22e4K4tnj7ckBcCLJJp+d+d6yGmi+BhVGQalQEDz6UYmOwq+S8OzdjTHeMm0EInYJYwJ0iKc9F1hTRJYExBel+sRiLzjs2G1etIQS06LB3BTUWED0R7bmWBvBTZmdqdMAhCXpgaluH4zto3Rwe44N3m65y4nruDo9QPNUpJCseTNfaauY3qtHQH8/weJjwiIWlOM7ugdm8+ika/uZBWC/qXpv/saUJGQteSLOMeC39KyPSqZ+YVC0AjiGHR5KlxoE0Chx7DU5WEhFyWUp4Ltv2+qUpigq2nlo0ErlM4ri3qaqsWR225QvJ1bBLtq0HErh6Yj2puHRIY9Q0MjM9Tw36Alam1tGJhNTFrloYeupuU0Tx9xaR8lL5JrKJtfp+5OZyYctocaD3j/5C3prIWLHufunLSIih5aNzWIL5FgGizqhtutpA/zCXWIyaew4IkBZWHV2w=
```
daryanypl: What's this do?

kfriedman: These are encrypted environment variables:
https://docs.travis-ci.com/user/encryption-keys/

LucasCioffi:
@kfriedman and @daryanypl I've made a few more changes, including adding an env file with encrypted values.  Travis can now successfully build the project.  I think it's worth another look @kfriedman just in case anything jumps out as incorrect.

Previously Travis was hanging on `npm test` and not completing.  I found that was because I set the `npm test` command to `jest --watchAll` which is useful for continuous testing when developing locally.  When I removed the `--watchAll` flag, Travis succeeded.

I expect that when we merge this into the Development branch, this will attempt a deploy to AWS.  I believe that all the necessary configurations are included.

kfriedman:
Looks right to me!

deploy_development.env
```diff
@@ -0,0 +1,15 @@
+#### #### #### ####
+# Encrypted Values
+#### #### #### ####
+
+CARD_CREATOR_USERNAME=AQECAHjqALewp8JBJNxIQvR4oY795dyG7INaGR1glMsTEgetggAAAHwwegYJKoZIhvcNAQcGoG0wawIBADBmBgkqhkiG9w0BBwEwHgYJYIZIAWUDBAEuMBEEDOwV7ZC8HGOJVmdzQgIBEIA5zcd1yYupstx/5f++p9g0GyxZc4Jw1/l3+mGa1ouTb1LYGyCaaO6DuLm0gfS0PZ/CZEmFWHKq4WW/
```
kfriedman: One thing I wanted to add:

Currently, I believe, there is no decryption functionality in the app. So, obviously, this will not work when deployed to development. We probably should create a separate task/tasks for this work. Just as a heads-up, there are other Node apps that decrypt environment variable so you might want to look to them as examples.

LucasCioffi: Sounds good @kfriedman I added it here: https://jira.nypl.org/browse/MLN-47

package.json
```diff
@@ -12,7 +12,7 @@
     "deploy-production": "TBD",
     "build-doc-deploy-package-qa": "npm run build-swagger-doc && npm run deploy-package-qa",
     "build-doc-deploy-package-production": "npm run build-swagger-doc && npm run deploy-package-production",
-    "test": "jest --watchAll",
+    "test": "jest",
```
daryanypl: Might you say your jest was... infinite?

kfriedman:
This could be the best comment ever!

LucasCioffi:
@EdwinGuzman @ktp242 @nonword Last week you were reviewing [PR #3](https://github.com/NYPL/dgx-patron-creator-service/pull/2) for using AWS KMS.  That was going to be merged into this branch for configuring Travis.

I've merged the two feature branches together so they only have to be reviewed once going forward.  I'd appreciate your review here.  Please let me know if you have any questions.

deploy_development.env
LucasCioffi: @kfriedman I've merged in the [feature branch related to AWS KMS](https://github.com/NYPL/dgx-patron-creator-service/pull/2).  The integration test is successfully interacting with the QA instance of Card Creator and the Patron Kinesis stream in the Sandbox environment.  Does that address the concern you raised in the comment above "no decryption functionality in the app"?

tests/integration/controllers/createPatron.test.js
```diff
@@ -0,0 +1,50 @@
...
+      "policy_type": "web_applicant",
+      "patron_agency": "198"
+    }
+  }
+}
+
+describe('createPatron v0.1 route', () => {
+  // TODO: Mocking the Kinesis stream as seen here: https://github.com/NYPL-discovery/node-nypl-streams-client/blob/pb/mocked-sdk-in-test-suite/test/encoding.test.js
+
+  if (process.env.INTEGRATION_TESTS == 'true') {
```
EdwinGuzman: Maybe this can go around the `describe` block instead of the individual tests? Not sure what will be easier when you add more tests.

deploy_development.env
kfriedman: This is great @LucasCioffi ! Thanks for your work on this.

tests/integration/controllers/createPatron.test.js
LucasCioffi: Thanks @EdwinGuzman I implemented your suggestion.

package.json
```diff
@@ -7,8 +7,10 @@
     "start": "swagger project start",
```
nonword: seems like `swagger` should be a dependency?

kfriedman: Are the Swagger modules providing any value? It seems like we have the Swagger docs in a file, right?

LucasCioffi: @nonword the original README suggested installing Swagger globally, `npm install swagger -g` but I'll change that and add it as a dependency.

@kfriedman the README suggests that we can edit the documentation with interactive UI, after running this command: `npm run swagger-edit`.  So that seems that there is some value.  Are you suggesting that we should or should not include Swagger as a dependency?

kfriedman: Yes, I guess I was wondering whether it should be a dependency. At the very least, I don't think it should be a production dependency. 

Also, I noticed `pg-promise` is a dependency. We don't use PostgreSQL anywhere, do we? 

It might be worth reviewing to make sure we're actually using all the dependencies.

nonword: True. Should probably be a dev-dependency. Seems like a handy util for running a server locally (and necessary for integration tests), but given that it's deployed as a lambda, no need for it as a production dependency. Lately I think we require all dependencies to be explicit, so the README suggestion that one install swagger globally is, I think, outdated.

LucasCioffi: @nonword ready for another look. Thank you!

LucasCioffi:
@nonword There are no further changes planned.  Ready for another look.  Thank you!

PR #3, opened 2018-06-22
Decrypting AWS environment variables

LucasCioffi:
We can merge this into the branch called "unit tests" so that Travis will work when we merge both into the development branch. 🤞


kfriedman:
Going to offer more through feedback but just wanted to say:

There is a page of Lambda Best Practices at:
https://docs.aws.amazon.com/lambda/latest/dg/best-practices.html

In particular, we want to make sure that we're caching/re-using expensive-ish calls like KMS:

>Take advantage of Execution Context reuse to improve the performance of your function. Make sure any externalized configuration or dependencies that your code retrieves are stored and referenced locally after initial execution. Limit the re-initialization of variables/objects on every invocation. Instead use static initialization/constructor, global/static variables and singletons. Keep alive and reuse connections (HTTP, database, etc.) that were established during a previous invocation.

Perhaps, this is getting re-used/cached between invocations but just wanted to drop a quick note and ask.

kfriedman:
Also, I'd suggest including someone with a bit more experience with Node/Lambdas in each PR for now. Maybe @EdwinGuzman, @ktp242, @nonword.

config/ccConfig.js
```diff
@@ -1,5 +1,31 @@
...
+        resolve(data.Plaintext.toString());
+      }
+    });
+  });
+};
+
 module.exports = {
-  username: process.env.CARD_CREATOR_USERNAME,
-  password: process.env.CARD_CREATOR_PASSWORD
+  username: decryptKMS(process.env.CARD_CREATOR_USERNAME),
```
nonword: Does `dotenv.config` assume a `.env` by default? Is one going to be available to travis?

Also `module.exports.username` is assigned `Promise` above - not a string value. The time we seem to use it is here: https://github.com/NYPL/dgx-patron-creator-service/blob/0758411d5c49a845bb49a21d3a017d462a0f4c62/api/controllers/createPatron.js#L113 Essentially we'll be sending `axios` a hash with `auth: { username: <Promise>, password: <Promise> }, ...`, right? I may be misreading something.

EdwinGuzman: That's right. It needs to be decrypted first before assigning the values to `username` and `password` and in this case it's returning a promise instead of the string values.

@LucasCioffi perhaps the kms decryption code should only go here, and in `createPatron` you call that decryption call and create this patron object there. You might want to call it directly before the axios call so that when the decryption is complete, then you can call axios on those values.

api/controllers/createPatron.js
```diff
@@ -8,6 +8,9 @@ const modelDebug = require('./../model/modelDebug.js');
 const modelStreamPatron = require('./../model/modelStreamPatron.js').modelStreamPatron;
 const streamPublish = require('./../helpers/streamPublish');
 
+var cardCreatorUsername;
```
EdwinGuzman: Use `let` instead of `var`.

config/ccConfig.js
```diff
@@ -1,5 +1,28 @@
...
+      if (err) {
+        reject(err);
+      } else {
+        resolve(data.Plaintext.toString());
+      }
+    });
+  });
+};
+
 module.exports = {
```
EdwinGuzman: The name of this file, since all this does is create and expose a function to decrypt kms strings, should have a more appropriate name.

package.json
```diff
@@ -38,18 +37,28 @@
...
     "pg-promise": "5.3.5",
     "promise": "7.1.1",
     "serverless": "1.9.0",
     "serverless-plugin-write-env-vars": "1.0.1",
+    "sinon": "^6.0.1",
     "swagger-express-mw": "0.6.0",
     "swagger-tools": "0.10.1"
   },
   "devDependencies": {
+    "assert": "^1.4.1",
```
EdwinGuzman: Can you remove the carets from the packages. We've found that at times a package may get a minor version bump which creates backwards compatibility issues.

tests/integration/controllers/createPatron.test.js
```diff
@@ -0,0 +1,40 @@
+const http = require('http')
```
EdwinGuzman: It doesn't look like you're using this package here.

PR #4, opened 2018-06-25
Adding unit tests to models

LucasCioffi:
Adding unit tests prior to adding v0.2 of the create patron route.
Jira ticket: https://jira.nypl.org/browse/MLN-38


ktp242:
Glad to see you added tests! In `package.json` there's a command to run `swagger project test`, which we never really use. I think we can delete it and also the related files to it. And add the new command line for running these new tests.

LucasCioffi:
Thanks @ktp242.  I removed the swagger default test files; ready for another look.

package.json
```diff
@@ -11,7 +11,7 @@
     "deploy-package-production": "node-lambda deploy -e production -f deploy_production.env",
     "build-doc-deploy-package-qa": "npm run build-swagger-doc && npm run deploy-package-qa",
     "build-doc-deploy-package-production": "npm run build-swagger-doc && npm run deploy-package-production",
-    "test": "swagger project test"
+    "test": "jest"
```
EdwinGuzman: Is `jest` the correct command or is it a typo for `test`?

tests/unit/models/modelDebug.test.js
```diff
@@ -0,0 +1,29 @@
...
+    expect(modelDebug.checkMissingRequiredField(filledFields)).toEqual([]);
+  });
+});
+
+describe('renderMissingFieldDebugMessage', () => {
+  it('creates a debug message if a field is missing', () => {
+    expect(modelDebug.checkMissingRequiredField(filledFields.concat(blankFields))).toEqual(blankFields);
+  });
+  it('does not create a debug message if there are no missing fields', () => {
+    expect(modelDebug.checkMissingRequiredField(filledFields)).toEqual([]);
```
EdwinGuzman: According to https://github.com/NYPL/dgx-patron-creator-service/blob/development/api/model/modelDebug.js#L23, this should return an empty array.

I think the bigger problem for the tests for `renderMissingFieldDebugMessage` is that it's not checking to see if the `Missing ${element.name}` string was added to the returned object.

tests/unit/models/modelRequestBody.test.js
```diff
@@ -0,0 +1,60 @@
+const modelRequestBody = require('../../../api/model/modelRequestBody');
```
EdwinGuzman: `modelSimplePatron` tests all the separate functions found in `modelRequestBody` which is fine, but I also think individual tests for the functions should also have tests, such as `updateDateOfBirthToBirthdate `.

package.json
LucasCioffi: Thanks for checking, @EdwinGuzman.  I'm using Facebook's Jest for JS testing: http://jestjs.io/ so although it looks like "jest" could be a typo, it is correct.

tests/unit/models/modelDebug.test.js
LucasCioffi: Great catch @EdwinGuzman.  The test for `renderMissingFieldDebugMessage` was calling the wrong method, so I fixed that.

tests/unit/models/modelRequestBody.test.js
LucasCioffi: Sounds good, @EdwinGuzman.  I'd be interested in a quick chat about this.  Coming from a Rails background, I follow Sandi Metz' philosophy of testing only the public methods; in a Node context, I interpret that as only the methods that are exported.  In that philosophy, the key is to test each branch of the code (including the private/non-exported methods) which is what I think I was able to do with the different variations of inputing birthday data.

This methodology gives a developer the flexibility to refactor or change the name of `updateDateOfBirthToBirthdate` while not requiring a rewrite of the tests as long as they didn't break any functionality.

But if our standard is to test the private/non-exported methods, then I certainly will!  

Since we're at the end of the sprint, I'd suggest we merge this (if everything else looks good) and we add a new ticket for tests on private methods next sprint.

ktp242:
The PR looks fine with me! I think it is a good start point.

Just a suggestion as Edwin pointed out, you might want to include in `REAMDME.md` or an new individual documentation file, to let the user know they have to install `jest` globally (If this is what you want them to do), or include jest in `package.json`, so when the users clone and install the repo, they won't run into errors.

And I think we should also update `CHANGELOG.md` for writing down this exciting new feature!!!

EdwinGuzman:
That's interesting. I'm not sure why that method of using private functions like that was done for this. I generally prefer to be able to test all functions and I have not encountered that in Rails before (because I don't have a Rails background). So, this seems fine.

But you should update the test command to include an internal `jest` package (like `"test": "/node_modules/.bin/jest"` or similar).

LucasCioffi:
Thanks, @EdwinGuzman. `npm test` works locally.  I've installed Jest as a dev dependency.  Is `"test": "/node_modules/.bin/jest"` still necessary?

EdwinGuzman:
No, it works fine as is now. Thanks.

Just curious, why did you choose jest?

LucasCioffi:
@EdwinGuzman I'm following some tutorials by [Mosh](https://learn.pluralsight.com/) and this is what he said about Jasmine, Mocha, and Jest:
```
*Jasmine:* "It's the earliest framework of these three."
*Mocha:* "Most popular on NPM but by design doesn't include features that are available in other testing libraries, so we need to use it alongside popular plugins like Chai and Sinon."
*Jest:* "A wrapper around Jasmine, (so if you know Jasmine, you know Jest), plus a code coverage tool.  Built by FB to test their React applications."
```
His tutorials use Jest, so that's what I chose to use.

PR #5, opened 2018-07-06
Remove some trailing commas for Node v6.10.3

LucasCioffi:
We were caught between a lint and a hard place.

The Airbnb linter required commas at the end of each list item, and that worked fine locally on Node 10.  When Travis deployed to AWS Lambda, that caused a problem, because our Lambda instance is running Node 6.10.

Running Node 6.10.3 locally, I was able to debug and resolve this problem.

Here is the original error on Lambda:
screen shot 2018-07-06 at 10 30 33 am

### DIFF
--- a/api/controllers/createPatron.js
+++ b/api/controllers/createPatron.js
@@ -75,8 +75,8 @@ function createPatron(req, res) {
           'invalid-request',
           'Missing required patron information.',
           null,
-          { form: ['Can not find the object "simplePatron".'] },
-        ),
+          { form: ['Can not find the object "simplePatron".'] } // eslint-disable-line comma-dangle
+        ) // eslint-disable-line comma-dangle
       ));
 
     return;
@@ -97,8 +97,8 @@ function createPatron(req, res) {
           'invalid-request',
           'Missing required patron information.',
           null,
-          debugMessage,
-        ),
+          debugMessage // eslint-disable-line comma-dangle
+        ) // eslint-disable-line comma-dangle
       ));
 
     return;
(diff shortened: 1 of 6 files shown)


kfriedman:
Wow, I'm surprised this is such an issue with Node 6.x. Seems to affect Node 8.x also.

One thing we might want to talk about adding to our testing strategy - that we've done in the best - is using node-lambda for mock events. I threw up an example PR that shows how to do it:

https://github.com/NYPL/dgx-patron-creator-service/pull/6

I'm not sure how this fits in with the existing CD process but it might be work exploring.

Thanks for your work on this!

api/controllers/createPatron.js
```diff
@@ -166,11 +166,11 @@ function createPatron(req, res) {
             req,
             res,
             statusCode,
-            modelResponse.errorResponseData(responseObject),
+            modelResponse.errorResponseData(responseObject) // eslint-disable-line comma-dangle
           );
         } else {
           renderResponse(req, res, 500, modelResponse.errorResponseData(
-            collectErrorResponseData(null, '', '', '', ''),
+            collectErrorResponseData(null, '', '', '', '') // eslint-disable-line comma-dangle
```
EdwinGuzman: Do you need all these comments here?

LucasCioffi: @EdwinGuzman the comments help the linter ignore the errors (ie missing comma).  I think that when we can get the error down to 0 errors, it's easier for us as developers to notice when there is a different error.

I'll chat about it with you today to see if there's a better way.

With the commas in, Node v6 has a problem.
With the commas out, the linter has a problem.
So I've removed the commas and added comments for the linter.

PR #6, opened 2018-07-06
Example of using test event with node-lambda

kfriedman:
Not suggesting we necessarily merge this... just wanted to show this to you. Let's talk more!

### DIFF
--- a/tests/events/request-docs.json
+++ b/tests/events/request-docs.json
@@ -1,0 +1,57 @@
+{
+  "body": "",
+  "resource": "/{proxy+}",
+  "requestContext": {
+    "resourceId": "123456",
+    "apiId": "1234567890",
+    "resourcePath": "/{proxy+}",
+    "httpMethod": "POST",
+    "requestId": "c6af9ac6-7b61-11e6-9a41-93e8deadbeef",
+    "accountId": "123456789012",
+    "identity": {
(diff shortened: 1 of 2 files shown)


tests/events/request-docs.json
```diff
@@ -0,0 +1,57 @@
...
+      "user": null,
+      "cognitoIdentityPoolId": null,
+      "cognitoIdentityId": null,
+      "cognitoAuthenticationProvider": null,
+      "sourceIp": "127.0.0.1",
+      "accountId": null
+    },
+    "stage": "prod"
+  },
+  "queryStringParameters": {
```
daryanypl: do these get filled out in later branches?  (sorry, not sure how jest works)

LucasCioffi:
@kfriedman what should we look for in a response aside from a 200 status?
screen shot 2018-07-23 at 4 09 53 pm

tests/events/request-docs.json
kfriedman: I guess it depends what you're testing. In this case, it's just testing the `docs/patron-creator` route with no query parameters.

kfriedman:
@LucasCioffi I think a 200 response is a good start and, if you wanted, a string like `swagger` just to make sure it actually did what you would expect?

PR #8, opened 2018-07-11
Lc/create patron v0.2

LucasCioffi:
This PR creates the v0.2 route for creating a patron.

If you hit `localhost:3001/api/v0.2/patrons` with the following JSON:
```
{
  "names": [
  "TestFirstName, TestLastName"
  ],
  "barcodes": [
    "barcode_2018_06_14_328pm1"
  ],
  "pin": "4316",
  "expirationDate": "2019-01-01",
  "birthDate": "1978-01-01",
"emails": [
        "test_email_2018_07_10_0951_a1@test.com",
        "test_email_2018_07_10_0951_b2@test.com"
    ],
  "patronType": 151,
  "patronCodes": {
    "pcode1": "s",
    "pcode2": "f",
    "pcode3": 5
  },
  "blockInfo": {
    "code": "-"
  },
  "addresses": [{
    "lines": [
      "ADDRESS LINE 1",
      "ADDRESS LINE 2"
    ],
  "type": "a"
  }],
  "phones": [{
    "number": "917-123-4567",
    "type": "t"
  }]
}
```
You will get back this response:
```
{
    "data": [
        {
            "id": 7130184,
            "names": [
                "TestFirstName, TestLastName"
            ],
            "barcodes": [
                "barcode_2018_06_14_328pm1"
            ],
            "expirationDate": "",
            "birthDate": "1978-01-01",
            "emails": [
                "test_email_2018_07_10_0951_a1@test.com",
                "test_email_2018_07_10_0951_b2@test.com"
            ],
            "pin": "4316",
            "patronType": 151
        }
    ],
    "count": 1,
    "totalCount": 0,
    "statusCode": 200,
    "debugInfo": []
}
```


LucasCioffi:
@EdwinGuzman and @ktp242 it would be great to hear your opinion about where this could be improved.

This publishes to the Kinesis stream called "NewPatron".  Sierra returns an error if we include pcode4 in the API request when creating a patron (until after the upgrade for a few months), so for now I'm removing it and hard-coding it as -1 so that we can at least meet the requirements of the schema associated with the stream.

ktp242:
It is unfortunate that we have to wait for Sierra.... bummer.
I guess we can document it in README for now. And not sure how MyLibraryNYC does project management, but you can also create a ticket and put it in the backlog as a reminder. 

Hope these approaches are helpful for you!

api/controllers/v0.2/createPatron.js
```diff
@@ -0,0 +1,193 @@
...
+
+  clientKey = process.env.CACHED_CLIENT_KEY ||
+    awsDecrypt.decryptKMS(process.env.ILS_CLIENT_KEY);
+  clientPassword = process.env.CACHED_CLIENT_PASSWORD ||
+    awsDecrypt.decryptKMS(process.env.ILS_CLIENT_SECRET);
+
+  Promise.all([clientKey, clientPassword]).then((values) => {
+    [process.env.CACHED_CLIENT_KEY, process.env.CACHED_CLIENT_PASSWORD] = values;
+
+    const username = process.env.CACHED_CLIENT_KEY;
```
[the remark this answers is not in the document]
LucasCioffi: On line 112 `process.env.CACHED_CLIENT_KEY` and `process.env.CACHED_CLIENT_PASSWORD` are overwritten with the returned values of the promises.

api/controllers/v0.2/createPatron.js
```diff
@@ -0,0 +1,191 @@
...
+    type: type || '',
+    message: message || '',
+    title: title || '',
+    debug_message: debugMessage || '',
+  };
+}
+
+/**
+ * renderResponse(req, res, status, message)
+ * Render the response from the ILS API.
```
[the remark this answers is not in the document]
LucasCioffi: This is for rendering a JSON response to the API consumer, the consumer that made the original request to create the patron.

api/controllers/v0.2/createPatron.js
```diff
@@ -0,0 +1,193 @@
...
+                streamPublish.streamPublish(
+                  process.env.PATRON_SCHEMA_NAME_V02,
+                  process.env.PATRON_STREAM_NAME_V02,
+                  streamPatron // eslint-disable-line comma-dangle
+                ).then((streamResponse) => { // eslint-disable-line no-unused-vars
+                  renderResponse(req, res, 201, modeledResponse);
+                  console.log('Published to stream successfully!'); // eslint-disable-line no-console
+                }).catch((streamError) => {
+                  renderResponse(req, res, 201, streamError);
+                  console.error(`Error publishing to stream: ${streamError}`); // eslint-disable-line no-console
```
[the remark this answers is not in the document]
LucasCioffi: @daryanypl I'm assuming yes, since that's the way it was before.  I think it can't hurt.

api/models/v0.2/modelDebug.js
```diff
@@ -0,0 +1,38 @@
...
+ * @param {array} array
+ * @return {array}
+ */
+function checkMissingRequiredField(array) {
+  return filter(array, element => (!element.value || isEmpty(element.value)));
+}
+
+/**
+ * renderDebugMessage(missingFields)
+ * Renders the debug message object based on the missing required fields.
```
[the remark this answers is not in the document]
LucasCioffi: @daryanypl I picked up the pattern from v0.1 so I suggest we put in a new ticket to address the wording later.

api/models/v0.2/modelStreamPatron.js
```diff
@@ -0,0 +1,39 @@
...
+   */
+  transformPatronRequest(requestBody, modeledResponse) {
+    return new Promise((resolve, reject) => {
+      const mergedPatronData = Object.assign({}, requestBody, modeledResponse.data[0]);
+      for (var key in mergedPatronData) {
+        if (modelStreamPatron.data.hasOwnProperty(key)) {
+          modelStreamPatron.data[key] = mergedPatronData[key];
+        }
+      }
+      modelStreamPatron.data.patronCodes.pcode4 = -1; // TODO: remove once pcode4 is not deleted in v0.2 of createPatron
```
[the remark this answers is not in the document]
LucasCioffi: @daryanypl yes that is the goal here.  We have to have a pcode4 in order to stream the data, so I put in -1 here.  The ticket to remove pcode4 hardcoding to -1 after Sierra is upgraded is here: https://jira.nypl.org/browse/MLN-86

api/swagger/swagger.yaml
```diff
@@ -137,79 +129,81 @@ definitions:
         example: true
       patron_id:
         type: string
-        example: "6367028"
+        example: '6367028'
```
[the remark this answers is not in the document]
LucasCioffi: @daryanypl that ID has been in there for over a year.  Source: https://github.com/NYPL/dgx-patron-creator-service/blame/development/README.md#L104

When I looked up this patron ID with the III client for the test environment, I see a test user with an email address of "digitallionprojectteam@nypl.org".

api/controllers/v0.1/createPatron.js
```diff
@@ -105,13 +106,13 @@ function createPatron(req, res) {
...
-  cardCreatorUsername = cardCreatorUsername ||
+  cardCreatorUsername = process.env.CACHED_CARD_CREATOR_USERNAME ||
     awsDecrypt.decryptKMS(process.env.CARD_CREATOR_USERNAME);
-  cardCreatorPassword = cardCreatorPassword ||
+  cardCreatorPassword = process.env.CACHED_CARD_CREATOR_PASSWORD ||
     awsDecrypt.decryptKMS(process.env.CARD_CREATOR_PASSWORD);
 
   Promise.all([cardCreatorUsername, cardCreatorPassword]).then((values) => {
-    [cardCreatorUsername, cardCreatorPassword] = values;
+    [process.env.CACHED_CARD_CREATOR_USERNAME, process.env.CACHED_CARD_CREATOR_PASSWORD] = values;
```
EdwinGuzman: Are these actually environment variables or just variables that you're using for local caching? If they're environment variables, you're okay with overriding them here?

api/controllers/v0.2/createPatron.js
```diff
@@ -0,0 +1,193 @@
...
+    return;
+  }
+
+  clientKey = process.env.CACHED_CLIENT_KEY ||
+    awsDecrypt.decryptKMS(process.env.ILS_CLIENT_KEY);
+  clientPassword = process.env.CACHED_CLIENT_PASSWORD ||
+    awsDecrypt.decryptKMS(process.env.ILS_CLIENT_SECRET);
+
+  Promise.all([clientKey, clientPassword]).then((values) => {
+    [process.env.CACHED_CLIENT_KEY, process.env.CACHED_CLIENT_PASSWORD] = values;
```
EdwinGuzman: Same comment as the above environment variables for these here.

api/controllers/v0.2/createPatron.js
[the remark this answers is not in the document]
EdwinGuzman: From my comments above, this is fine but they don't have to be environment variables. It seems like it'll work just fine if they're local variables.

config/deploy_development.env
```diff
@@ -1,16 +1,31 @@
...
+#      Encrypted Values for API v0.1
+# ---- ---- ---- ---- ---- ---- ---- ---- ---- ---- ----
 CARD_CREATOR_USERNAME=AQECAHjqALewp8JBJNxIQvR4oY795dyG7INaGR1glMsTEgetggAAAHwwegYJKoZIhvcNAQcGoG0wawIBADBmBgkqhkiG9w0BBwEwHgYJYIZIAWUDBAEuMBEEDOwV7ZC8HGOJVmdzQgIBEIA5zcd1yYupstx/5f++p9g0GyxZc4Jw1/l3+mGa1ouTb1LYGyCaaO6DuLm0gfS0PZ/CZEmFWHKq4WW/
 CARD_CREATOR_PASSWORD=AQECAHjqALewp8JBJNxIQvR4oY795dyG7INaGR1glMsTEgetggAAAJ4wgZsGCSqGSIb3DQEHBqCBjTCBigIBADCBhAYJKoZIhvcNAQcBMB4GCWCGSAFlAwQBLjARBAwcjlURKNSKTli7ZxYCARCAVyilOh9U/x5dJWW3t52Dsy1MG/NLPm4bjWCuVqNCcf1p7yLZuUHF2Lh7qoa/qCqzxNGfm+IbCb01IEt+G6TNf0qEMx9sa66OtWUnJiRQPSL2G8YGJ3lK5Q==
 
 
-#### #### #### ####
-# Unencrypted Values
-#### #### #### ####
+#      Encrypted Values for API v0.2
```
EdwinGuzman: These are okay to have in this file and be committed to the repo?...

api/controllers/v0.2/createPatron.js
```diff
@@ -0,0 +1,193 @@
...
+
+    const username = process.env.CACHED_CLIENT_KEY;
+    const password = process.env.CACHED_CLIENT_PASSWORD;
+    const basicAuth = `Basic ${base64(`${username}:${password}`)}`;
+
+    axios.post(process.env.ILS_CREATE_TOKEN_URL, {}, {
```
EdwinGuzman: What happens if the `ILS_CREATE_TOKEN_URL` and `ILS_CREATE_PATRON_URL` are both empty for whatever reason? There's no check to make sure they're not empty to then proceed to the API call.

api/controllers/v0.2/createPatron.js
```diff
@@ -0,0 +1,193 @@
...
+      headers: {
+        'Content-Type': 'application/json',
+        Authorization: basicAuth,
+      },
+    })
+      .then((tokenResponse) => {
+        axios.post(process.env.ILS_CREATE_PATRON_URL, patronData, {
+          headers: {
+            'Content-Type': 'application/json',
+            Authorization: `Bearer ${tokenResponse.data.access_token}`,
```
EdwinGuzman: This is also assuming that you got a successful API call and return value AND that there's an `access_token` in `data` from the returned value.

api/models/v0.2/modelDebug.js
```diff
@@ -0,0 +1,38 @@
+const filter = require('underscore').filter;
+const isEmpty = require('underscore').isEmpty;
```
EdwinGuzman: Does the `import { filter, isEmpty } from 'underscore';` syntax work here? Just putting out another option for syntax but not a big deal.

api/models/v0.2/modelResponse.js
```diff
@@ -0,0 +1,98 @@
...
+ *
+ * @param {object} data
+ * @param {number} status
+ * @return {object}
+ */
+function modelPatronCreatorResponse(responseData, status, requestData) {
+  return {
+    data: [
+      {
+        id: parseInt(responseData.link.split('/').pop()),
```
EdwinGuzman: Can you add `10` as the radix value to `parseInt` just for an extra safety catch?

api/swagger/swaggerDoc.json
```diff
@@ -1 +1,261 @@
...
+---
+swagger: '2.0'
```
EdwinGuzman: This is not json anymore so the file should be renamed.

EdwinGuzman:
Looks fine but just general comments (and some that should be updated like the file name).

api/models/v0.2/modelResponse.js
```diff
@@ -0,0 +1,98 @@
...
+        names: requestData.names,
+        barcodes: requestData.barcodes,
+        expirationDate: '',
+        birthDate: requestData.birthDate,
+        emails: requestData.emails,
+        pin: requestData.pin,
+        patronType: parseInt(requestData.patronType),
```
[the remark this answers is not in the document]
LucasCioffi: Fortunately, no.  If Sierra returns an error (ie a status code that isn't in the 200's) then we won't get down this path.

The error is caught with `.catch((response) => {` in `api/controllers/v0.2/createPatron.js`

api/controllers/v0.1/createPatron.js
LucasCioffi: @EdwinGuzman I need a better way to cache these per-user rather than at the environment level.  I was definitely wrong in putting them at the process.env level.

api/models/v0.2/modelDebug.js
LucasCioffi: If I overwrite lines 1 and 2 with `import { filter, isEmpty } from 'underscore';` I see `SyntaxError: Unexpected token import` when running the tests.

Apparently this isn't supported by default yet: https://stackoverflow.com/questions/39436322/node-js-syntaxerror-unexpected-token-import

api/models/v0.2/modelResponse.js
LucasCioffi: Sounds good; I did that here and a few lines below in the other `parseInt`

LucasCioffi:
@EdwinGuzman Thanks for all your comments.  I addressed them in the latest commit.

LucasCioffi:
I think this is OK @ktp242; I'll drop by to discuss.

Something new that we're doing with patron creator route v0.2 is that we are sending barcode and expiration date.  Sierra is saving them through the integration tests as I can see in this screenshot.
screen shot 2018-07-18 at 10 54 51 am

PR #10, opened 2018-07-19
Lc update swagger

LucasCioffi:
We had to add `write:patron` to the `api_auth` part of the YAML file and update the test data for the v0.2 route to match the format of the data we expect (ie no more SimplePatron).


LucasCioffi:
@nnat425 @daryanypl You can test this at http://localhost:3001/docs
The scope of this PR was to add write:patron and to get the test data correct so that dev-platformdocs would work for the MLN app's credentials.
I don't think you can submit data via the Swagger docs locally, though.  That can be another ticket.

nnat425:
I don't think its possible to actually test this scope change on http://localhost:3001 change till the codebase get's pushed to the development environment (dev-platformdocs). The reason for that is because we need the API Gateway in our infrastructure to test this change which it only lives in development and not our actual local.

kfriedman:
I think I'm a little confused. The Express `/docs/patron-creator` route reads `swaggerDoc.json` and the file updated in this PR is `swagger.yaml`. Which also isn't a YAML file anymore. I have a feeling we should strip out the Swagger dependency to prevent confusion and remove an unnecessary dependency. I created a PR (#11) from this branch that I think we should consider merging into this PR.

nnat425:
@LucasCioffi I'm still getting back a 401 unauthorized.

nnat425:
@LucasCioffi never mind, looks good now. However, I'm not getting the expected response. Most likely due to the Sierra Test API upgrade going on.

PR #12, opened 2018-07-21
Add logging class

daryanypl:


LucasCioffi:
@daryanypl thanks for doing this!!

When I run `INTEGRATION_TESTS=true npm test` in the terminal where the app is running, I do not see a log message from the v0.1 route at api/controllers/v0.2/createPatron.js with `logger.debug('Published to stream successfully!', {routeTag: ROUTE_TAG});` but I do see "Published to log successfully!" from the v0.1 route at api/controllers/v0.1/createPatron.js where we are still using `console.log`.

Should I expect to still see a log message from both places?

Once it is working as intended, you or I can replace all the existing `console.log` calls with the new `logger.debug` method.

daryanypl:
To see the log messages, we have to set the environment logging level to "debug".  By default, it's set to "info".

PR #13, opened 2018-07-23
Update Swagger documentation

kfriedman:
This PR updates the Swagger docs:

- Simplifies the error response (i.e. no need to wrap in `data`; more like other Platform error responses)
- Fixes the error in `barcodes` so it's returned as an array
- Adds `id` in the response

### DIFF
--- a/api/swagger/swaggerDoc.json
+++ b/api/swagger/swaggerDoc.json
@@ -5,9 +5,9 @@
     "title": "dgx-patron-creator-service"
   },
   "host": "localhost:3001",
-  "basePath": "/",
+  "basePath": "/api",
   "schemes": [
-    "https"
+    "http"
   ],
   "consumes": [
     "application/json"
@@ -402,7 +402,7 @@
           "items": {
             "type": "string"
           },
-          "example": "barcode_2018_07_19_328pm"
+          "example": [ "barcode_2018_07_19_328pm" ]
         },
         "pin": {
           "type": "string",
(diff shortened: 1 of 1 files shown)


api/swagger/swaggerDoc.json
```diff
@@ -534,21 +534,15 @@
...
-        }
-      }
-    },
-    "PatronModelV02": {
       "required": [
         "names",
         "patronType"
       ],
       "properties": {
+        "id": {
```
daryanypl: Do we get/want to dictate the Sierra patron id?  Is this the promary key?  Would it be generated, if we didn't indicate it?

api/swagger/swaggerDoc.json
```diff
@@ -633,66 +630,6 @@
       }
     },
     "400ErrorResponseV02": {
-      "required": [
```
daryanypl: Is specifying too much detail on the error responses a no-no?

LucasCioffi:
@daryanypl I merged this in so I could fix merge conflicts with https://github.com/NYPL/dgx-patron-creator-service/pull/14

api/swagger/swaggerDoc.json
LucasCioffi: Kevin suggested removing `required` on any definitions related to responses, so that is taken care of in https://github.com/NYPL/dgx-patron-creator-service/pull/14

kfriedman: `required` is really intended more for the Request object (i.e. what parameters or fields are required) so it's a little strange (and maybe even misleading/inaccurate) to have it in the Response definition. That's why I suggested removing it.

PR #14, opened 2018-07-23
Handle Sierra and environment config errors better MLN-93

LucasCioffi:
New encrypted values for ILS_CLIENT_KEY and ILS_CLIENT_SECRET.

DRYing up the createPatron v0.2 method and improving error handling in that function, especially for when we cannot connect to Sierra.

Tests for some private methods in createPatron.js.

### DIFF
--- a/api/controllers/v0.1/createPatron.js
+++ b/api/controllers/v0.1/createPatron.js
@@ -6,6 +6,9 @@ const modelResponse = require('./../../models/v0.1/modelResponse.js');
 const modelDebug = require('./../../models/v0.1/modelDebug.js');
 const modelStreamPatron = require('./../../models/v0.1/modelStreamPatron.js').modelStreamPatron;
 const streamPublish = require('./../../helpers/streamPublish');
+const logger = require('../../helpers/Logger');
+
+const ROUTE_TAG = 'CREATE_PATRON_0.1';
 
 let cardCreatorUsername;
 let cardCreatorPassword;
@@ -134,7 +137,7 @@ function createPatron(req, res) {
           ))
           .then(() => {
             renderResponse(req, res, 201, modeledResponse);
-            console.log('Published to stream successfully!'); // eslint-disable-line no-console
+            logger.debug('Published to stream successfully!', { routeTag: ROUTE_TAG });
           })
           .catch((error) => {
             renderResponse(req, res, 201, modeledResponse);
@@ -159,12 +162,10 @@ function createPatron(req, res) {
             response.response.data.debug_message // eslint-disable-line comma-dangle
           );
 
-          const statusCode = (responseObject.status) ? responseObject.status : 500;
-
           renderResponse(
             req,
             res,
-            statusCode,
+            (responseObject.status || 500),
             modelResponse.errorResponseData(responseObject) // eslint-disable-line comma-dangle
           );
         } else {
--- a/api/controllers/v0.2/createPatron.js
+++ b/api/controllers/v0.2/createPatron.js
@@ -5,37 +5,59 @@ const modelResponse = require('./../../models/v0.2/modelResponse.js');
 const modelDebug = require('./../../models/v0.2/modelDebug.js');
 const modelStreamPatron = require('./../../models/v0.2/modelStreamPatron.js').modelStreamPatron;
 const streamPublish = require('./../../helpers/streamPublish');
-const logger = require('../../helpers/Logger')
+const logger = require('../../helpers/Logger');
+const encode = require('../../helpers/encode');
+const customErrors = require('../../helpers/errors');
 
-const ROUTE_TAG = "CREATE_PATRON_20";
-
-
-function base64(string) {
-  return Buffer.from(string).toString('base64');
-}
-
+const ROUTE_TAG = 'CREATE_PATRON_0.2';
 let ilsClientKey;
 let ilsClientPassword;
+let ilsToken;
+let ilsTokenTimestamp;
 
 /**
- * collectErrorResponseData(status, type, message, title, debugMessage)
- * Model the response from a failed request.
+ * validateEnvVariable(envVariableName)
+ * Validate that a specific environment variable is present.
  *
- * @param {number} status
- * @param {string} type
- * @param {string} message
- * @param {string} title
- * @param {string} debugMessage
- * @return {object}
+ * @param {HTTP response} res
+ * @param {string} envVariableName
+ * @throws {InvalidEnvironmentConfiguration}
  */
-function collectErrorResponseData(status, type, message, title, debugMessage) {
-  return {
-    status: status || null,
-    type: type || '',
-    message: message || '',
-    title: title || '',
-    debug_message: debugMessage || '',
-  };
+function validateEnvVariable(res, envVariableName) {
+  const envVariable = process.env[envVariableName];
+
+  if (!envVariable || isEmpty(envVariable)) {
+    throw new customErrors.InvalidEnvironmentConfiguration(`${envVariableName} was not set.`);
+  }
+}
+
+/**
+ * validateEnvironmentAndRequest(req, res)
+ * Validate the request format and environment variables related to the ILS and streaming service.
+ *
+ * @param {HTTP request} req
+ * @param {HTTP response} res
+ */
+function validateEnvironmentAndRequest(req, res) {
+  const envVariableNames = [
+    'ILS_CLIENT_KEY', 'ILS_CLIENT_SECRET', 'SCHEMA_API_BASE_URL', 'ILS_CREATE_PATRON_URL',
+    'ILS_CREATE_TOKEN_URL', 'PATRON_STREAM_NAME_V02', 'PATRON_SCHEMA_NAME_V02',
+  ];
+  envVariableNames.forEach((envVariableName) => {
+    validateEnvVariable(res, envVariableName);
+  });
+
+  if (!req.body || isEmpty(req.body)) {
+    throw new customErrors.InvalidRequest('The request body is empty.');
+  }
+  // Check if we get all the required information from the client
+  const requiredFields = [];
+  const missingFields = modelDebug.checkMissingRequiredField(requiredFields);
+
+  if (missingFields.length) {
+    const debugMessage = modelDebug.renderMissingFieldDebugMessage(missingFields);
+    throw new customErrors.InvalidRequest(debugMessage);
+  }
 }
 
 /**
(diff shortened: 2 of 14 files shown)


api/swagger/swaggerDoc.json
```diff
@@ -174,10 +174,6 @@
     },
     "PatronsCreatorResponseV01": {
       "type": "object",
-      "required": [
```
LucasCioffi: @kfriedman I removed the required fields in this file as you suggested.

api/controllers/v0.2/createPatron.js
```diff
@@ -55,74 +55,154 @@ function renderResponse(req, res, status, message) {
 }
 
 /**
- * createPatron(req, res)
- * The callback for the route "/patrons".
- * It will fire a POST request to the ILS API for creating a new patron.
+ * streamPatron(streamPatron, modeledResponse)
```
ktp242: I assume the parameters will be like `(req, res, streamPatron, modeledResponse)`? If this is the case, line 58 should be `streamPatron(req, res, streamPatron, modeledResponse)`

api/controllers/v0.2/createPatron.js
```diff
@@ -55,74 +55,154 @@ function renderResponse(req, res, status, message) {
...
 
-    return;
-  }
+/**
+ * callAxiosToCreatePatron(req, res, tokenResponse)
+ * Stream the new patron data to Kinesis
+ *
+ * @param {HTTP request} req
+ * @param {HTTP response} res
+ * @param {object} req.body
```
ktp242: it should be `tokenResponse`?

api/controllers/v0.2/createPatron.js
```diff
@@ -55,74 +55,154 @@ function renderResponse(req, res, status, message) {
...
+        ));
+      }
+    });
+}
+
+/**
+ * validateEnvVariable(envVariableName)
+ * Validate that a specific environment variable is present.
+ *
+ * @param {string} envVariableName
```
ktp242: Also this should have `res`

ktp242:
It looks good! But I just noticed that we have `config/deploy_development.env` pushed. I think we probably should keep it discreet?

ktp242:
Thanks for the explanation.

tests/unit/models/v0.2/modelResponse.test.js
```diff
@@ -43,53 +43,45 @@ const ilsResponse = {
 };
 
 const patronCreatorResult = {
-  data: [
```
LucasCioffi: We are intentionally removing the containing array so that we only return one patron record.

config/deploy_development.env
```diff
@@ -6,8 +6,8 @@ CARD_CREATOR_PASSWORD=AQECAHjqALewp8JBJNxIQvR4oY795dyG7INaGR1glMsTEgetggAAAJ4wgZ
 
 #      Encrypted Values for API v0.2
 # ---- ---- ---- ---- ---- ---- ---- ---- ---- ---- ----
-ILS_CLIENT_KEY=AQECAHjqALewp8JBJNxIQvR4oY795dyG7INaGR1glMsTEgetggAAAHoweAYJKoZIhvcNAQcGoGswaQIBADBkBgkqhkiG9w0BBwEwHgYJYIZIAWUDBAEuMBEEDJPuxivxfMCssbQcQAIBEIA3dgSMfy540wG69KJECgAiMq6DVVi0bPJwi7nXJ8LI/I37MMGzONmQjapdWJv17ceMzSm5gEeCBA==
-ILS_CLIENT_SECRET=AQECAHjqALewp8JBJNxIQvR4oY795dyG7INaGR1glMsTEgetggAAAIYwgYMGCSqGSIb3DQEHBqB2MHQCAQAwbwYJKoZIhvcNAQcBMB4GCWCGSAFlAwQBLjARBAzPb80eDfdT9AMQ9ucCARCAQlxE/9PXuDhq0teK1IdZQxjmcf9bW2mfxEdT38yrPfnxk4mRCYwl9HL9iug9dnzH3KpwumsFFgrflcVQk1Hd1m7jEA==
+ILS_CLIENT_KEY=AQECAHjqALewp8JBJNxIQvR4oY795dyG7INaGR1glMsTEgetggAAAHoweAYJKoZIhvcNAQcGoGswaQIBADBkBgkqhkiG9w0BBwEwHgYJYIZIAWUDBAEuMBEEDJSW/4wv/vPtqNFgbgIBEIA3bFZJLvQqiqhLkRyo2bkFo7yly2cNTD+H831fwZKzBHSEgLwI3VcJz+5HFdly1EppqXZIgjX2NA==
```
LucasCioffi: These are encrypted values which had to be updated after last week's Sierra upgrade.

api/models/v0.2/modelResponse.js
```diff
@@ -76,14 +68,12 @@ function parseTypeURL(str) {
  */
 function modelErrorResponseData(obj) {
   return {
-    data: {
```
LucasCioffi: We are intentionally removing the containing array so that we only return one patron record.

api/models/v0.2/modelResponse.js
```diff
@@ -10,26 +10,18 @@ const url = require('url');
  */
 function modelPatronCreatorResponse(responseData, status, requestData) {
   return {
-    data: [
```
LucasCioffi: We are intentionally removing the containing array so that we only return one patron record.

api/controllers/v0.2/createPatron.js
```diff
@@ -163,73 +261,52 @@ function createPatron(req, res) {
...
-            }
-          });
       });
   });
 }
 
-module.exports = {
-  createPatron,
-};
+if (process.env.NODE_ENV === 'test') {
```
LucasCioffi: This is unconventional, but I wanted to export `validateEnvironmentAndRequest` and `validateEnvVariable` in the test environment so that I could test them.

api/controllers/v0.2/createPatron.js
```diff
@@ -163,73 +261,52 @@ function createPatron(req, res) {
       .then((tokenResponse) => {
         if (!tokenResponse.data || !tokenResponse.data.access_token) {
           reject();
+          return;
+        }
+        callAxiosToCreatePatron(req, res, tokenResponse);
+      })
+      .catch((tokenErrorResponse) => {
```
LucasCioffi: I added this new `.catch` block.

LucasCioffi:
@nonword I added some comments to explain why I made certain changes.  If you have time to review this today or tomorrow, I would greatly appreciate it.

api/controllers/v0.2/createPatron.js
```diff
@@ -135,25 +216,42 @@ function createPatron(req, res) {
...
+function createPatron(req, res) {
+  if (validateEnvironmentAndRequest(req, res) === false) { return false; }
 
   ilsClientKey = ilsClientKey || awsDecrypt.decryptKMS(process.env.ILS_CLIENT_KEY);
   ilsClientPassword = ilsClientPassword || awsDecrypt.decryptKMS(process.env.ILS_CLIENT_SECRET);
 
+  // eslint-disable-next-line max-len
+  // Delete pcode4 data to prevent an error on Sierra TODO: remove this line once pcode4 is accepted; also remove related line in v0.2 modelStreamPatron.js
+  if (req.body && req.body.patronCodes) {
+    delete req.body.patronCodes.pcode4;
```
nonword: Just want to make sure I understand this. So sierra patron api fails if you send it a pcode4, so you're removing pcode4 from the posted data and plan to submit another PR in the future to remove these places where you're removing `pcode4`?

api/controllers/v0.2/createPatron.js
```diff
@@ -55,74 +55,155 @@ function renderResponse(req, res, status, message) {
...
+      const modeledResponse =
+        modelResponse.patronCreator(response.data, response.status, req.body);
+      modelStreamPatron.transformPatronRequest(
+        req.body, modeledResponse // eslint-disable-line comma-dangle
+      )
+        .then((streamPatronData) => {
+          streamPatron(req, res, streamPatronData, modeledResponse);
+        })
+        .catch((error) => {
+          renderResponse(req, res, 201, modeledResponse);
```
nonword: What kind of error would be caught here? Seems like an error writing to the stream would be caught inside `streamPatron`, so this is only catching errors thrown by `modelStreamPatron.transformPatronRequest`, which should resolve almost unconditionally. If there are known types of errors caught here, is there a reason you'd want to render `modeledResponse` with http status 201 in those cases? In other words, we can only reach this `catch` block if an error was thrown in `modelStreamPatron.transformPatronRequest`, so it's odd to communicate that the patron record was created.

api/controllers/v0.2/createPatron.js
```diff
@@ -55,74 +55,155 @@ function renderResponse(req, res, status, message) {
...
+  streamPublish.streamPublish(
+    process.env.PATRON_SCHEMA_NAME_V02,
+    process.env.PATRON_STREAM_NAME_V02,
+    streamPatronData // eslint-disable-line comma-dangle
+  ).then((streamResponse) => { // eslint-disable-line no-unused-vars
+    renderResponse(req, res, 201, modeledResponse);
+    console.log('Published to stream successfully!'); // eslint-disable-line no-console
+  }).catch((streamError) => {
+    renderResponse(req, res, 201, streamError);
+    console.error(`Error publishing to stream: ${streamError}`); // eslint-disable-line no-console
```
nonword: Is this the right http status code for the condition where the patron was created but there was an error writing to the stream? Maybe it is; I'm not sure. But it may be confusing to send a different payload for the same 201 status code - `streamError` instead of `modeledResponse`

Also, Is this an opportunity to centralize error handling by returning the promise without a `catch` call so that the caller has to handle the error? (i.e. allow it to be caught by the `catch` on what is now line 104).

api/controllers/v0.2/createPatron.js
```diff
@@ -55,74 +55,155 @@ function renderResponse(req, res, status, message) {
...
+  const envVariableNames = [
+    'ILS_CLIENT_KEY', 'ILS_CLIENT_SECRET', 'SCHEMA_API_BASE_URL', 'ILS_CREATE_PATRON_URL',
+    'ILS_CREATE_TOKEN_URL', 'PATRON_STREAM_NAME_V02', 'PATRON_SCHEMA_NAME_V02',
+  ];
+  envVariableNames.forEach((envVariableName) => {
+    if (validateEnvVariable(res, envVariableName) === false) {
+      return false;
+    }
+    return true;
+  });
```
nonword: `validateEnvVariable` as written validates by writing directly to the HTTPResponse object. I don't think there's any utility checking the return value of `validateEnvVariable` as written. In any case, returning from within a `forEach` callback has no effect here. Is the desire that you write a `400` response for the first missing ENV var and then stop processing?

api/controllers/v0.2/createPatron.js
```diff
@@ -55,74 +55,155 @@ function renderResponse(req, res, status, message) {
...
+          collectErrorResponseData(null, '', '', '', '') // eslint-disable-line comma-dangle
+        ));
+      }
+    });
+}
 
-  if (!process.env.ILS_CREATE_PATRON_URL || isEmpty(process.env.ILS_CREATE_PATRON_URL)) {
+/**
+ * validateEnvVariable(envVariableName)
+ * Validate that a specific environment variable is present.
```
nonword: I think this needs more explanation. *How* does it validate a specific env var? I would expect it to throw an error if the env var is missing/invalid - or just return `false` - but this seems to write directly to the HTTPResponse object (implying that no further processing is necessary/expected). It doesn't look like the caller is checking to see if the env var validated. In effect, I think this function may call `res.status(500)...` for each missing env variable in series.

api/controllers/v0.2/createPatron.js
```diff
@@ -163,73 +261,52 @@ function createPatron(req, res) {
...
+        if (tokenErrorResponse.response && tokenErrorResponse.response.data) {
+          const responseObject = collectErrorResponseData(
+            tokenErrorResponse.response.status,
+            'invalid-request',
+            `${tokenErrorResponse.message} from ILS.`,
+            'ILS Response',
+            tokenErrorResponse.response.data // eslint-disable-line comma-dangle
+          );
+
+          const statusCode = (responseObject.status) ? responseObject.status : 500;
```
nonword: Unnecessary ternary. could be `const statusCode = responseObject.status || 500;`

api/controllers/v0.2/createPatron.js
```diff
@@ -135,25 +216,42 @@ function createPatron(req, res) {
...
+  }
+
   Promise.all([ilsClientKey, ilsClientPassword]).then((values, reject) => {
     [ilsClientKey, ilsClientPassword] = values;
 
     const username = ilsClientKey;
     const password = ilsClientPassword;
     const basicAuth = `Basic ${base64(`${username}:${password}`)}`;
-
     axios.post(process.env.ILS_CREATE_TOKEN_URL, {}, {
```
nonword: This may be beyond the scope of this work, but seems a shame to fetch a new token for every request. Ideally we'd use previously fetched tokens until they expire (and then refresh them). (But again, this may be beyond the scope of this work)

api/controllers/v0.2/createPatron.js
LucasCioffi: Yes, there will be a new PR.  Sierra fails if pcode4 is included and the stream fails if pcode4 is not included.  Publishing to the stream happens afterwards so the sequence is to remove and then add.

api/controllers/v0.2/createPatron.js
```diff
@@ -55,74 +55,155 @@ function renderResponse(req, res, status, message) {
...
+    renderResponse(req, res, 201, streamError);
+    console.error(`Error publishing to stream: ${streamError}`); // eslint-disable-line no-console
+  });
+}
 
-    return;
-  }
+/**
+ * callAxiosToCreatePatron(req, res, tokenResponse)
+ * Stream the new patron data to Kinesis
```
daryanypl: This is the same doc as for the streamPatron method?  What's the diff?

api/controllers/v0.2/createPatron.js
```diff
@@ -55,74 +55,155 @@ function renderResponse(req, res, status, message) {
...
+
+        renderResponse(
+          req,
+          res,
+          statusCode,
+          modelResponse.errorResponseData(responseObject) // eslint-disable-line comma-dangle
+        );
+      } else {
+        renderResponse(req, res, 500, modelResponse.errorResponseData(
+          collectErrorResponseData(null, '', '', '', '') // eslint-disable-line comma-dangle
```
daryanypl: Is it reasonable to send an error response with no data inside?  (I didn't trace all the code logic, so am not sure if this is your absolute last resort.)

api/models/v0.2/modelResponse.js
```diff
@@ -10,26 +10,18 @@ const url = require('url');
...
-        blockInfo: requestData.blockInfo,
-        addresses: requestData.addresses,
-        phones: requestData.phones,
-      },
-    ],
-    count: 1,
-    totalCount: 0,
-    statusCode: status,
-    debugInfo: [],
+    id: parseInt(responseData.link.split('/').pop(), 10),
```
daryanypl: If parseInt fails here or on parsing requestData.patronType, then modelPatronCreatorResponse will crash.  Is this the best course of action, or would a try/catch block be called for?

api/controllers/v0.2/createPatron.js
LucasCioffi: @daryanypl I'm picking up the established pattern from v0.1.  If it should change, let's make this a tech debt card on Jira.

api/controllers/v0.2/createPatron.js
LucasCioffi: Righto!

api/controllers/v0.2/createPatron.js
daryanypl: ok with me

LucasCioffi:
@ktp242 @nonword
Thank you for your suggestions this week. This PR is ready for another look.  There were some substantial refactors based on both of your suggestions.

We have existing integration tests and one unit test for the createPatron method, and a ticket for more unit tests are in the backlog to verify all the different error paths.

api/helpers/Logger.js
```diff
@@ -13,70 +13,68 @@ const nyplLogLevels = {
...
+          result.meta = JSON.stringify(options.meta);
         }
 
-        return JSON.stringify(result)
-      }
-    })
-  )
+        return JSON.stringify(result);
+      },
+    }) // eslint-disable-line comma-dangle
```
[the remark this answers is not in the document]
LucasCioffi: The comment `// eslint-disable-line comma-dangle` is necessary to get past the linter if we don't have a comma there.  If we do have a comma there, `npm start` throws this error unfortunately:
```
/Users/lucascioffi/Desktop/apps/dgx-patron-creator-service/api/helpers/Logger.js:68
  );
  ^
SyntaxError: Unexpected token )
```
So having the comment seems necessary.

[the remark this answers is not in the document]
ktp242: I see. Looks like airbnb override the `ignore` from native Eslint config. I think the parameters of a function don't need a dangling coma though especially when it breaks the app. Do you want to just set the `ignore` back? So you don't need the comment.

https://github.com/airbnb/javascript/issues/1169

api/controllers/v0.1/createPatron.js
```diff
@@ -134,7 +137,7 @@ function createPatron(req, res) {
           ))
           .then(() => {
             renderResponse(req, res, 201, modeledResponse);
-            console.log('Published to stream successfully!'); // eslint-disable-line no-console
+            logger.debug('Published to stream successfully!', { routeTag: ROUTE_TAG });
```
daryanypl: :))

api/controllers/v0.2/createPatron.js
```diff
@@ -5,17 +5,76 @@ const modelResponse = require('./../../models/v0.2/modelResponse.js');
...
+  if (!envVariable || isEmpty(envVariable)) {
+    throw new customErrors.InvalidEnvironmentConfiguration(`${envVariableName} was not set.`);
+  }
 }
 
-let ilsClientKey;
-let ilsClientPassword;
+/**
+ * validateEnvironmentAndRequest(req, res)
+ * Validate the environment variables and the request.
```
daryanypl: This is a bit vague.  Which env vars, validate for what, etc.?

api/controllers/v0.2/createPatron.js
```diff
@@ -5,17 +5,76 @@ const modelResponse = require('./../../models/v0.2/modelResponse.js');
...
+let ilsToken;
+let ilsTokenTimestamp;
 
+/**
+ * validateEnvVariable(envVariableName)
+ * Validate that a specific environment variable is present.
+ *
+ * @param {HTTP response} res
+ * @param {string} envVariableName
+ * @throws {InvalidEnvironmentConfiguration}
```
daryanypl: 👍

api/controllers/v0.2/createPatron.js
```diff
@@ -29,6 +88,14 @@ let ilsClientPassword;
  * @return {object}
  */
 function collectErrorResponseData(status, type, message, title, debugMessage) {
+  logger.error(
+    `status_code: ${status}, ` +
```
daryanypl: Does node have a way to build strings with interpolating, rather than concatenating?  Is it generally considered safer?

api/models/v0.2/modelResponse.js
daryanypl: Do we know which patron record gets chosen if there's more than one?

api/controllers/v0.2/createPatron.js
LucasCioffi: @daryanypl yes Node has a way.  Each of these lines is doing interpolation and then all the lines are concatenated.

We are writing to a log on AWS rather than a database so I think the security considerations are negligible.  I don't know of a security difference between concatenation vs. interpolation.

api/models/v0.2/modelResponse.js
LucasCioffi: @daryanypl Sierra will respond with the one patron we created, even if there are others with duplicate emails or other data.

LucasCioffi:
@daryanypl to answer the question you asked, the ILS will return the one patron that was just created.

api/models/v0.2/modelResponse.js
```diff
@@ -1,5 +1,21 @@
...
+ * @return {number || string}
+ */
+function tryCatchParseInt(string) {
+  try {
+    return parseInt(string, 10);
+  } catch (error) {
+    return 'unparseable';
+  }
+}
+
```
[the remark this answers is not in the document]
nonword: As written, you'll never hit line 16 above because `parseInt` does not ever throw anything:

```
> [undefined, null, '', '1', 'a', 3.14, true, false, [], {}, 1, new Object(), new Error()].map(parseInt)
[ NaN, NaN, NaN, 1, NaN, 3, NaN, NaN, NaN, NaN, 1, NaN, NaN ]
```

There's no reason to wrap `parseInt` in a try-catch. For that matter, there's no reason to run the regex test if you're just going to return 'NaN' when it fails, because `parseInt` is going to do that in most cases for you. (Technically it returns the global `NaN`, but when cast to a string it will be `"NaN"`.)

More to the original point, what error condition are you anticipating here, and should you really allow patron records to be created with `id` equal to "NaN"? Will that lead to a db schema type error for that column? Is that a condition in which you should fail record creation?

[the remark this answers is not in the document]
LucasCioffi: @nonword after chatting with you, I think we need to send the data that we have to the app that called our service.  That means we continue responding with a 201, because that's the status we got back from Sierra, and we should include a message "No link or ID from ILS" so that we can more easily debug the stream.

```
axios.post(process.env.ILS_CREATE_PATRON_URL, req.body, {
    headers: {
      'Content-Type': 'application/json',
      Authorization: `Bearer ${ilsToken}`,
    },
  })
    .then((axiosResponse) => {
      const modeledResponse = modelResponse.patronCreator(axiosResponse.data, axiosResponse.status, req.body); // eslint-disable-line max-len
      modelStreamPatron.transformPatronRequest(req.body, modeledResponse)
        .then((streamPatronData) => {
          streamPatron(req, res, streamPatronData, modeledResponse);
        })
        .catch(() => {
          // eslint-disable-next-line max-len
          renderResponse(req, res, 201, modeledResponse); // respond with 201 even if streaming fails
        });
    })
    .catch((axiosErrorResponse) => {...
```

PR #22, opened 2018-08-15
Remove hardcoding of pcode4 MLN-190

LucasCioffi:
The MLN app is going to start sending accurate pcode4 values to this service once https://github.com/NYPL/MyLibraryNYCApp/pull/39/ is merged in.

With this PR, we are now forwarding that to the ILS.


api/models/v0.2/modelStreamPatron.js
```diff
@@ -9,7 +9,12 @@ const modelStreamPatron = {
     emails: [],
     pin: null,
     patronType: null,
-    patronCodes: null,
+    patronCodes: {
```
nnat425: So in the other pull request, https://github.com/NYPL/dgx-patron-creator-service/pull/21/files, you set each pcode to be null but in this PR you're setting it to 0. Can you explain why in one PR they all are equaling null and in another, they're 0?

LucasCioffi: For consistency, I made the change to `null` just now.  Thanks @nnat425

nnat425: Nice.

PR #37, opened 2020-03-31
Adding git branch workflow and updating the README.

EdwinGuzman:
Adding a PR template file for future PRs. Let me know what you think of it and if it's missing anything.

Updated the README a bit but added a new section on how to contribute based on the branches that are there. ~A big TODO is to verify that Travis does deploy a new lambda version automatically. If that's true, then we can update the "Deployment" section.~

I can verify that Travis _did_ deploy a new version based of the `development` branch in the NYPL Sandbox account. I'll update the README accordingly because I'm confident that's how QA and Production work as well.
Screen Shot 2020-03-31 at 1 03 41 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 #### Update
 
 - Updated npm packages and added Prettier for code formatting.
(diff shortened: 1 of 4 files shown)


EdwinGuzman:
Updated.

PR #46, opened 2020-04-22
Import models from ruby on rails Card Creator

EdwinGuzman:
## Description

Resolves [DQ-284](https://jira.nypl.org/browse/DQ-284) and partly resolves [DQ-286](https://jira.nypl.org/browse/DQ-286).

~This was going fine but then got out of hand with the size of the code that was being brought over. So, breaking this up into different PRs.~ It's also a bit difficult because some models (284) are dependent on an ILS connector (286). That means not everything is fully implemented but the skeleton is there. Note that the receiving branch is `v0.3` so we can make updates until it's ready for the main branch pipeline.

Some parts also depend on the Service Objects implementation which should now be easier for you to connect to and see what will be calling it and how.

## Motivation and Context

Decommissioning the Card Creator and bringing in that code here.

## How Has This Been Tested?

Just unit test for now, which are also not extensive until more of the models and other classes are complete.

## Checklist:


api/controllers/v0.3/AddressValidationAPI.js
```diff
@@ -0,0 +1,26 @@
+/**
+ * A class that uses Service Objects to validate addresses.
+ * TODO: Finish the implementation.
+ */
+const AddressValidationApi = () => {
+  this.ALTERNATE_ADDRESSES_TYPE = 'alternate-addresses';
```
adriana-alter, line 6: What are alternate addresses--potential canonicalized versions of the address the user put in, or multiple residences, or...?

api/controllers/v0.3/ILSHelper.js
```diff
@@ -0,0 +1,390 @@
...
+      message: {
+        search_string: query,
+        username: process.env["ILS_USERNAME"],
+        password: process.env["ILS_PASSWORD"],
+      },
+    });
+  }
+
+  create_patron(patron) {
+    if (!patron.validForIls) {
```
adriana-alter, line 76: What does it mean for a patron to be valid for ILS--can we get a comment?

api/controllers/v0.3/ILSHelper.js
```diff
@@ -0,0 +1,390 @@
...
+  // Internal: Format card fields for ILS patron creation/updates.
+  formattedPatronData(patron) {
+    let email;
+    let birthdate;
+    let workAddress;
+    if (patron.email) {
+      email = patron.email.gsub(/\s+/, "").toUpperCase;
+    }
+    if (patron.birthdate) {
+      // birthdate = patron.birthdate.strftime("%Y%m%d000000");
```
adriana-alter, line 132: Was this line supposed to be removed...?

api/controllers/v0.3/ILSHelper.js
```diff
@@ -0,0 +1,390 @@
...
+    }
+
+    let address = this.formatAddress(patron.address);
+    if (patron.worksInCity) {
+      workAddress = this.formatAddress(patron.workAddress);
+    }
+
+    let fullName = patron.name.toUpperCase();
+    if (!fullName.includes(",")) {
+      // Existing TODO: Replace this code with a call to a dedicated human name
```
adriana-alter, line 143: There doesn't seem to be a branch for what happens if the name does include a comma.  (Also, why would someone's name ever include a comma??)

api/controllers/v0.3/ILSHelper.js
```diff
@@ -0,0 +1,390 @@
...
+
+    let fields = [
+      fieldData["name"],
+      fieldData["address"],
+      fieldData["ptype"],
+      fieldData["username"],
+      fieldData["pin"],
+      fieldData["ecommunications_pref"],
+      // notice_pref_as_field,
+      // home_lib_as_field,
```
adriana-alter, line 214: Were these supposed to be uncommented/removed?

api/controllers/v0.3/AddressValidationAPI.js
EdwinGuzman: Let's say you put in your address as "6th Avenue", the ILS can return an alternate address of "Avenue of the Americas"

api/controllers/v0.3/ILSHelper.js
EdwinGuzman: Right, so many of this is still the ruby to js conversion and, like the description says, it's skeleton code without updating too much of the logic. In this case, the string needs to be converted but it does it a few times so it might be better as a utility function which I haven't gotten to yet.

api/controllers/v0.3/ILSHelper.js
EdwinGuzman: That can be found in the Card model code (Card and Patron are interchangeable in the ruby code and makes this confusing). But it's valid for the ILS if the patron data is valid and there's a ptype. I'll write more comments.

api/controllers/v0.3/ILSHelper.js
EdwinGuzman: Maybe the user added it by mistake or the ILS returned "Guzman, Edwin"... not sure. Most of this file has the existing logic so I haven't really touched most of the branching logic. That's in the next PR.

api/controllers/v0.3/ILSHelper.js
EdwinGuzman: I have a better idea of how this works and much of this can be removed. I didn't understand how those values were created in ruby so just commented it out for now.

api/controllers/v0.3/ILSHelper.js
```diff
@@ -0,0 +1,390 @@
...
+    return {
+      patronFields: {
+        patron_field: fields,
+      },
+      patronID: patron.patronId,
+    };
+  }
+
+  blankField() {
+    return { fieldTag: "", marcTag: 0, value: "", value_is_binary: false };
```
adriana-alter, line 250: When would you want this?  Comment?

api/controllers/v0.3/ILSHelper.js
```diff
@@ -0,0 +1,390 @@
...
+    // TODO convert date object
+    let expiration = time.strftime("%Y%m%d000000");
+    return this.customField(
+      IlsHelper.EXPIRATION_FIELDTAG,
+      expiration,
+      IlsHelper.STRING_MARCTAG
+    );
+  }
+
+  longExpirationAsField(patron) {
```
adriana-alter, line 271: Is there a reason to keep this and `shortExpirationAsField` as separate functions (instead of passing an argument for standard/temporary)?

api/controllers/v0.3/ILSHelper.js
```diff
@@ -0,0 +1,390 @@
...
+  // If the field requires the STRING_MARCTAG, it must be included in the
+  // IlsHelper.WITH_MARK_TAG array.
+  //
+  // Signature
+  //
+  //   <field>_as_field([value])
+  //
+  // field - a field name
+  //
+  // Returns a Hash representing the custom field.
```
adriana-alter, line 302: I assume this was referring to a Ruby hash (i.e. a JS object)...?

api/controllers/v0.3/ILSHelper.js
```diff
@@ -0,0 +1,390 @@
...
+        : null;
+      return this.customField(fieldTag, value, marcTag);
+    } catch (e) {
+      // throw new NotImplementedError(`//${field} field details have not been set`);
+      throw new Error(`//${field} field details have not been set`);
+    }
+  }
+}
+
+IlsHelper.MINOR_AGE = 11;
```
adriana-alter, line 320: It might not be worth changing this, if it's used in a bunch of different places...but maybe call this `CHILD_AGE` or something.  It's confusing having "minor" mean "under 12."

api/controllers/v0.3/ILSHelper.js
EdwinGuzman: The previous code had a lot of "every function should only do one specific thing". Is that normal ruby convention as well? In any case, yes, there's lot of optimizations we can do but still reading the code and trying to make sure that's possible to do. There are some cases where many functions aren't even used, and some cases where margin will be harder than expected.

api/controllers/v0.3/ILSHelper.js
EdwinGuzman: Ah, yes. Those I'll update by removing.

api/controllers/v0.3/ILSHelper.js
EdwinGuzman: That's true but it could also be convention used by the ILS so I'll double-check with the ILS team first. It might make it easier to make calls if we use the convention the ILS expects rather than having to convert and remember both those variable names.

api/controllers/v0.3/ILSHelper.js
EdwinGuzman: Not sure, but just copied it over. This whole file will be updated soon.

api/models/v0.3/modelAddress.js
```diff
@@ -0,0 +1,145 @@
...
+        "Address lines must be less than 100 characters combined";
+      return false;
+    }
+    this.is_valid = true;
+    return true;
+  }
+
+  str_to_bool(str) {
+    if (!str) {
+      return undefined; // Is this right? prev was nil
```
adriana-alter, line 39: Should probably be `null`

api/models/v0.3/modelCard.js
```diff
@@ -0,0 +1,343 @@
...
+      const birthDate = new Date(card.birthdate);
+      const age = today.getFullYear() - birthDate.getFullYear();
+      const m = today.getMonth() - birthDate.getMonth();
+      if (m < 0 || (m === 0 && today.getDate() < birthDate.getDate())) {
+        age = age - 1;
+      }
+
+      if (min_age > age) {
+        // TODO acquire an appropriate error message here for below minimum_age.
+        card.errors["age"] = ["Date of birth is below the minimum age of 13."];
```
adriana-alter, line 84: Is it actually supposed to be 13 here but 11 in the other place?

api/controllers/v0.3/ILSHelper.js
adriana-alter: Atomic functions are normal Ruby convention, but...this is pretty extreme.

api/models/v0.3/modelPolicy.js
```diff
@@ -0,0 +1,140 @@
...
+        id: IlsHelper.WEB_APPLICANT_PTYPE,
+        desc: IlsHelper.PTYPE_TO_TEXT["WEB_APPLICANT_PTYPE"],
+      },
+    },
+    card_type: {
+      standard: IlsHelper.WEB_APPLICANT_EXPIRATION_TIME,
+      temporary: IlsHelper.WEB_APPLICANT_EXPIRATION_TIME,
+    },
+    required_fields: ["birthdate"],
+    minimum_age: 13,
```
adriana-alter, line 136: And again:  11 or 13?

EdwinGuzman: Right, I don't know. I'm keeping those values for now to also show the inconsistencies in the older codebase and we'll correct them once it's settled (but I do think the policy says 13).

EdwinGuzman:
This should be all the models but ignore the ilsHelper file for now. That uses the old SOAP based API and we'll be using the new REST API, so that whole file will be a new PR.

api/controllers/v0.3/UsernameValidationAPI.js
```diff
@@ -1,19 +1,18 @@
...
   // class IntegrationError < StandardError; end
-  const USERNAME_PATTERN = /\A[a-zA-Z0-9]{5,25}\z/;
+  const USERNAME_PATTERN = /^[a-zA-Z0-9]{5,25}$/;
   const AVAILABLE_USERNAME_TYPE = "available-username";
   const UNAVAILABLE_USERNAME_TYPE = "unavailable-username";
   const INVALID_USERNAME_TYPE = "invalid-username";
 
-  const STANDARD_CARD_TYPE = Card.STANDARD_CARD_TYPE;
-  const TEMPORARY_CARD_TYPE = Card.TEMPORARY_CARD_TYPE;
+  const STANDARD_CARD_TYPE = "standard";
```
adriana-alter, line 14: Why the shift away from the `Card` constants?

EdwinGuzman: Didn't want the cyclical dependencies. The Card model imports this class.

api/models/v0.3/modelPolicy.js
```diff
@@ -1,140 +1,134 @@
...
-  uses_an_approved_policy() {
-    const keys = Object.keys(Policy.ils_policy);
-    if (!keys.include(this.policy_type)) {
-      this.errors.push(`${policy_type} is invalid, must be ${keys.join(", ")}`);
+  const usesAnApprovedPolicy = () => {
+    const keys = Object.keys(ilsPolicy);
+    if (!keys.includes(policyType)) {
+      console.log(
+        `${policyType} policy type is invalid, must be of type ${keys.join(
+          ', ',
```
adriana-alter, line 108: Why the extra comma?

EdwinGuzman: It's not extra, it's used to join the keys in the `ilsPolicy` object.

adriana-alter: The first comma (the one in the single quotes, with the space after it) is joining the keys.  But then there's a second one...?

EdwinGuzman: Ah, I see. It's `eslint`'s dangling comma rule. Usually, it's done in objects and it is a bit weird seeing it here.

api/controllers/v0.3/ILSHelper.js
EdwinGuzman: I'll update this whole file in a different PR since it is its own ticket. This file doesn't affect the other files since it is mocked in tests.

EdwinGuzman:
Updated with more documentation for the classes and functions. For now, the `IlsHelper` and `Barcode` files are temporary. They need their own tickets since they need to call the ILS and the `Barcode` class requires a db.

EdwinGuzman:
Also, I'd love to move this to Typescript which will help improve readability, type checking, and remove the not-very-reliable esdoc param strings.

api/models/v0.3/modelAddress.js
```diff
@@ -98,18 +147,28 @@ class Address {
...
-  // TODO: Need to call validation API to validate the address
-  // which depends on Service Objects
-  validationresponse(isWorkAddress = undefined) {
+  /**
+   * validationResponse(isWorkAddress)
+   * TODO: Need to call validation API to validate the address
+   * which depends on Service Objects
+   *
+   * @param {boolean} isWorkAddress
+   */
```
adriana-alter, line 156: AAAH I didn't realize I was blocking you.  I'll take care of it next sprint.  Sorry about that.

api/models/v0.3/modelCard.js
```diff
@@ -49,6 +56,15 @@ export const CardValidator = () => {
     }
   };
 
+  /**
+   * validateAddress(card, addressType, workAddress)
+   * Returns the card object with updated validated address or errors based
+   * on policy and ILS verification.
+   *
+   * @param {Card object} card
+   * @param {string} addressType - "address" or "workAddres"
```
adriana-alter, line 65: Small typo ("workAddres**s**")

api/models/v0.3/modelCard.js
```diff
@@ -134,6 +161,12 @@ class Card {
...
 
+  /**
+   * validate()
+   * Runs simple validations to make sure that the arguments are present and
+   * passes the needed requirements, and once those pass, then validates the
+   * card's address and username against the ILS.
+   */
   validate() {
     const validateByPolicy = ["email", "birthdate"];
     // These four values are necessary
```
adriana-alter, line 172: FOUR values?  Email, birthdate, and...?

api/models/v0.3/modelCard.js
```diff
@@ -181,6 +223,11 @@ class Card {
     );
   }
 
+  /**
+   * checkValidUsername()
+   * If the current usernamename has been validated or not, return that value.
```
adriana-alter, line 228: This is a little confusing...

api/models/v0.3/modelCard.js
```diff
@@ -189,47 +236,97 @@ class Card {
...
+   * live in NY state but work in NYC.
+   */
   cardDenied(address, isWorkAddress) {
     if (this.policy.policy.serviceArea) {
       // If they are not in NY state, check to see if the patron
       // works in NYC.
       if (!address.inState(this.policy)) {
         if (isWorkAddress) {
-          // && !address.inCity(this.policy)) {
+          // && workAddress.inCity(this.policy)) {
```
adriana-alter, line 329: Why is this commented out...?

api/models/v0.3/modelAddress.js
EdwinGuzman: Well, once you create the validation APIs with Service Objects, you'd then implement this function, too.

api/models/v0.3/modelCard.js
EdwinGuzman: The comment is for the next line but I can write them down again on this line.

api/models/v0.3/modelAddress.js
adriana-alter: Cool, I'll stop feeling guilty now.

api/models/v0.3/modelCard.js
EdwinGuzman: Forgot to remove. I'll remove.

EdwinGuzman:
Nudging @adriana-alter .

adriana-alter:
@EdwinGuzman I thought you were still making changes!  If not, go ahead and merge.

EdwinGuzman:
I thought you were still reviewing all the files. If it all makes sense, approve it.

PR #47, opened 2020-04-24
Models pt2

EdwinGuzman:
This is a continuation of #46 because I didn't want it to be so big. With all the things you noted in the previous PR, it might make more sense for me to merge this branch `models-pt2` into the original `models` PR and so you can review everything. Do you think that will be helpful for you?

I'm thinking that'll be the way to go since updates in this PR fixes concerns in the other PR. It'll be much bigger in size, but I think it'll be easier to read. Let me know what you think.


EdwinGuzman:
Note that some validation api files are mocked/empty since it involves the Service Objects work you'll be doing. Hopefully, the skeleton will make it easier to see how those classes should be used (along with the flow from the existing ruby code).

adriana-alter:
Yes, seems better to merge them (even though it's going to be HUGE!).

PR #48, opened 2020-04-30
Ils api

EdwinGuzman:
## Description

Resolves [DQ-286](https://jira.nypl.org/browse/DQ-286).

This adds an IlsClient class that has two main functions, `validate` and `createPatron`. The `validate` function validates barcodes or usernames against the ILS and the `createPatron` creates a patron in the ILS. 

A couple of PR notes: 
* There's a larger lambda-related architectural question about where is the right place to store URLs and the token so I'm leaving that off for now.
* There's no error handling or returning error responses. To not make this bigger than it already is, I'm leaving that off for another ticket/PR. That said, this version returns the correct response and acceptable error responses that will be enhanced in the next iteration.
* The file `api/controllers/v0.3/createPatron.js` has a lot of commented out and boilerplate code. That file will be fixed in a different ticket and PR. BUT, you can still use this to test out making requests. I suggest you use Postman for testing it out  using the endpoint `localhost:3001/api/v0.3/patrons` and a mocked patron request object like :
```
{
	"name": "Edwin Guzman",
	"address": {
		"line1": "476 5th Ave",
		"city": "New York City",
		"state": "NY",
		"zip": "10018",
		"isResidential": "true"
	},
	"birthdate": "01/01/1988",
	"pin": "1234",
	"username": "edwinguzman",
	"ecommunicationsPref": "true"
}
```
* The `ecommunications` field isn't working when attempting to send to the ILS and I will talk to the ILS team about what's the right way to send it. I did add a comment for that in the code. 

## How Has This Been Tested?

Yes, there are multiple branches of logic for creating patrons and checking availability. I'm liking mocking using jest because it's very easy to do. Still, you can do end-to-end testing by running the app in development mode (I can give you ILS test credentials) and using Postman to make requests.

## Checklist:


adriana-alter:
The code looks good. But the Postman test is resulting in "Cannot GET /api/v0.3/patrons".

EdwinGuzman:
It's a POST request.
I also don't think you have ILS keys to test end-to-end so I'll send that separately.

PR #49, opened 2020-05-04
Error handling

EdwinGuzman:
## Description

This resolves [DQ-289](https://jira.nypl.org/browse/DQ-289).

This adds better error handling for ILS requests but also bad cases in the rest of the models. Instead of returning boolean values for validation or requests, it now propagates the error that is thrown and the main express route function handles catching it and responding with the correct response. A new response from the ILS is a 409 which is a server error from them meaning that the barcode/username is a duplicate and so not available. If the ILS is down or not returning a 500 error, then we send a 502 meaning that our server isn't the problem but the problem happened somewhere else down the line.
Example:
Screen Shot 2020-05-04 at 5 05 53 PM

**Note**: The `createPatron` file still isn't complete and is a separate ticket, but I'm continuously making updates there because it's easier to test end-to-end using Postman. Try playing around with the request to get some of the following errors, but not everything is working well yet.

Missing required value, a 400 for a bad request from the user:
Screen Shot 2020-05-04 at 6 00 25 PM

Wrong pin, a 400 for a bad request from the user:
Screen Shot 2020-05-04 at 6 00 37 PM

And one end-to-end is when trying to create a patron, if the username isn't valid, that error is caught (from the ILS) and returned:
Screen Shot 2020-05-04 at 6 06 37 PM

## How Has This Been Tested?

Using Postman to call the endpoints. Check above.

## Checklist:


api/models/v0.3/modelCard.js
```diff
@@ -169,16 +169,22 @@ class Card {
    * passes the needed requirements, and once those pass, then validates the
    * card's address and username against the ILS.
    */
-  validate() {
+  async validate() {
+    // if (!this.ilsClient) {
```
EdwinGuzman, line 173: This is a TODO.

EdwinGuzman:
@adriana-alter  nudge

PR #51, opened 2020-05-05
Cleaning up the main endpoint function that express uses to route requests.

EdwinGuzman:
## Description

This resolves [DQ-285](https://jira.nypl.org/browse/DQ-285). This cleans up the main file that was used to call functions for the Express request routes. Note that no tests are included in this since it requires end-to-end tests and not unit tests. Some places where address validation should go are commented out for now.

## Motivation and Context

The main file was getting too long and when adding the /validations endpoints, there was a lot of redundancy. Refactored functions that validated the app's environment variables as well as calls to the ILS for token generation.

There are plenty of comments and hopefully, they help to understand the high-level flow of the requested data and the success/failure response.

Here's a rough outline of the process (which I'll include as documentation in a later PR):
* Make a request to `/api/v0.3/patrons` in front of the API Gateway to create a new patron
* The API Gateway routes the request to this Lambda function
* The lambda function checks to see if all environment variables are set for keys, passwords, URLs, and Kinesis stream names
* If everything is fine, then make a request to the ILS to get a token. That token is used to make authenticated calls to the API to create the patron and to validate usernames and barcodes.
* Once the token is obtained, create _one_ ilsClient with the token and proceed to validate usernames or barcodes or create a patron.

Since this is a lambda, we want to cache as much as possible before it gets terminated and a new lambda is created.

## How Has This Been Tested?

Using postman, check `localhost:3001/api/v0.3/patrons` with an object similar to 
```
{
	"username": "someusername",
	"name": "edwin",
	"address": {
		"line1": "476 5th Avenue",
		"city": "New York",
		"state": "New York",
		"zip": "10018",
		"isResidential": "true"
	},
	"pin": "1234",
	"birthdate": "01-01-1988"
}
```
Also, try a POST request to `localhost:3001/api/v0.3/validations/username` with an object like:
```
{
	"username": "someusername"
}
```

What you specifically want to see is request errors like:
Screen Shot 2020-05-05 at 5 48 11 PM

You can remove an environment variable from `/config/deploy_development.env` and it should  return a similar error:
Screen Shot 2020-05-05 at 5 47 07 PM

You can also mock that the IlsClient purposely throws an error to mock that the ILS is down:
Screen Shot 2020-05-05 at 5 47 35 PM

## Checklist:


EdwinGuzman:
@adriana-alter  nudge

EdwinGuzman:
@adriana-alter 👋

PR #52, opened 2020-05-07
Adding a postgres db and setting up the Barcode class to use it.

EdwinGuzman:
## Description

A WIP that resolves [DQ-292](https://jira.nypl.org/browse/DQ-292).  Tests are currently missing but wanted to get some feedback on generating the barcodes that get added to a new patron account request. A later part of this will be setting it up so it connects to an RDS instance on AWS, so some configs will be updated in a different PR.

@leonardr you don't have to check the code itself, but going over the logic in the `api/models/v0.3/modelBarcode.js` file would be helpful. I'm looking for the lowest numbered barcode that is unused first. If there are none, then it's getting the lowest numbered barcode that is used and subtracting one to get the next available barcode. It's checked against the ILS for availability. If the barcode is available, it's sent with the request to create a patron. If it's successful, then the barcode is added to the database and marked as used. If it's not successful, then it's not added to the database. If the barcode was already in the database and the patron was successfully created, then the `used` flag is updated to true.

## How Has This Been Tested?

Through local database and Postman to make requests.

Screen Shot 2020-05-07 at 4 38 04 PM

## Checklist:


api/models/v0.3/modelBarcode.js
```diff
@@ -1,65 +1,170 @@
...
+
+      // If the barcode is available, insert it into the database.
+      if (barcodeAvailable) {
+        if (newBarcode) {
+          this.addBarcode(barcodeToTry, true);
+        } else {
+          this.markUsed(barcodeToTry, true);
+        }
+        break;
+      }
```
leonardr, line 156: This seems vulnerable to a race condition where two processes both notice that a barcode is available on the ILS side. What happens if the INSERT statement issued by addBarcode fails?

api/models/v0.3/modelBarcode.js
```diff
@@ -1,65 +1,170 @@
...
+
+      // Otherwise, let's try once more with the next barcode in the sequence.
       tries -= 1;
+      barcodeToTry = `${barcodeToTry - 1}`;
```
leonardr, line 160: Antonio mentioned in the call yesterday that it's more complicated than this because the last digit of a barcode is a checksum digit.

So if the last used barcode is `233331234567895`, the first available barcode is not `233331234567894`, it's `23333123456788X`, where X is the result of the checksum algorithm.

This requirement went unstated last time, and it's caused a lot of problems, so I'm making a big deal out of it now. This should be a separate function. It's OK to implement the function as x-1 for now, but once you have the checksum algorithm it should implement that algorithm.

api/models/v0.3/modelCard.js
```diff
@@ -407,17 +418,28 @@ class Card {
...
 
-    // create the patron
-    const response = await this.ilsClient.createPatron(this);
+    // Now create the patron in the ILS.
+    try {
+      response = await this.ilsClient.createPatron(this);
+    } catch (error) {
+      // We want to catch the error here just to free up the barcode in the
+      // database, but continue to throw the same error as an API response.
+      await this.freeBarcode(this.barcode);
```
leonardr, line 440: I'm also concerned about a race condition here. Let's imagine two processes running at once, A and B.

They both get the idea that the next available barcode is 23333567.

Process A creates a patron with that barcode and marks the barcode as used.
Process B tries to create a patron with the same barcode and gets an error -- the patron already exists.
Process B then frees the barcode because it encountered an error.

Now it looks like 23333567 is free for some future process to use, even though there's a patron with that barcode.

leonardr:
Looking over this code, I remembered we were originally concerned with load on the ILS. I think that's why we relied so heavily on reserving barcodes ahead of time. This strategy -- to use reserved barcodes if any are available but to go out to the ILS if they're not -- is a good compromise.

That said, I can see a number of race conditions here. It's nothing we can't fix, but this stuff is difficult to test and there is middle ground between the old system (clearly broken) and fetching a new barcode every time.

Like, we could have a background script that runs every 5 minutes and makes sure there are 1000 free barcodes in the database. Combine that with better monitoring and I'd be comfortable with controller code that errors out (with a helpful message) if it turns out there are no free barcodes.

api/models/v0.3/modelBarcode.js
EdwinGuzman: Yeah, I just got the algorithm this morning and I'll implement it soon.

EdwinGuzman:
I think, from reading the ruby code, that even though there are reserved barcodes to use, each time a patron is about to be created, the barcode is checked for availability in the ILS. So it was free in the database but double-checked in the ILS every time. It seems like a lot of load on the ILS, but I kept this same pattern.

I like the background script but I'm not sure how to do that if this is a lambda. It might make more sense, now, to convert this from a lambda to an ec2.

api/controllers/v0.3/UsernameValidationAPI.js
```diff
@@ -58,13 +58,14 @@ const UsernameValidationAPI = (args) => {
    * @param {string} username
    */
   const usernameAvailable = async (username) => {
+    const isBarcode = false;
```
adriana-alter, line 61: Why is this needed?

api/controllers/v0.3/endpoints.js
```diff
@@ -283,7 +283,8 @@ async function createPatron(req, res) {
     });
 
   let address = new Address(req.body.address);
-  const policy = Policy({ policyType: "webApplicant" });
+  // TODO: What should the default policy be?
+  const policy = Policy({ policyType: req.body.policyType || "simplye" });
```
adriana-alter, line 287: What happened to "webApplicant"?

api/models/v0.3/modelBarcode.js
```diff
@@ -1,65 +1,170 @@
...
-  markUnused(barcode) {
-    noop();
+  /**
+   * getNextAvailableBarcode()
+   * Generates a new barcode in the database from the last one in the database,
+   * and verifies it's available in the ILS. If it is, return it.
+   */
+  async getNextAvailableBarcode() {
+    let barcode = await this.generateNewBarcode();
+    let tries = 3;
```
adriana-alter, line 19: Where does `tries` get decremented?

api/models/v0.3/modelCard.js
```diff
@@ -39,7 +40,7 @@ const CardValidator = () => {
       card.errors["username"] = ["Username is not available or valid"];
     }
 
-    if (card.email && !/\A[^@]+@[^@]+\z/.test(card.email)) {
+    if (card.email && !/^[^@]+@[^@]+$/.test(card.email)) {
```
adriana-alter, line 43: Why the change?

api/controllers/v0.3/UsernameValidationAPI.js
EdwinGuzman: The `available` function from the `ilsClient` can check barcode or username, but the flag is set so that it sets up the params correctly.

api/controllers/v0.3/endpoints.js
EdwinGuzman: It's still available but the requirement might change on what's the default.

api/models/v0.3/modelBarcode.js
EdwinGuzman: Good point.

api/models/v0.3/modelCard.js
EdwinGuzman: Js regex seems to be slightly different than the ruby version.

leonardr: Yeah, those look like old-time Perl regexes to me.

api/models/v0.3/modelCard.js
EdwinGuzman: I updated the database checking and setting of `used` based on the other comments. _I think_ the next commit might help fix this issue.

If the barcode is new, an error will be thrown when it's inserted into the database if it's already there and a new barcode will be used.

If the barcode is an existing unused barcode, then an error will be thrown when updating the value of `unused` to true. So if A uses the barcode it's set to true and if B tries to set it to true and it's already true, then it will try a new barcode.

EdwinGuzman:
The last commit can be reviewed but it doesn't implement the new barcode algorithm.

EdwinGuzman:
I added a file to create 14-digit barcodes using the luhn-algorithm. We're supposed to also support the 15-digit NYC ID but that shouldn't be stored, nor are we creating them. I'm assuming that the patron will somehow add their NYC ID as an added barcode (or just as their single barcode) through the UI and just bypass the barcode creating and go directly into the ILS request. But, I think that's something to tackle later since I also don't have much of the requirements or workflow around that.

EdwinGuzman:
I will set this "ready for review" once I write tests.

api/models/v0.3/modelBarcode.js
```diff
@@ -70,12 +71,18 @@ class Barcode {
...
         "select barcode from barcodes where used=true order by barcodes asc limit 1;";
       result = await db.query(query);
-      barcode = `${parseInt(result.rows[0].barcode, 10) - 1}`;
+      // Remove the last digit from the existing barcode since we need a
+      // 13-digit number to pass through the luhn-algorithm.
+      barcode = Math.floor(parseInt(result.rows[0].barcode, 10) / 10);
+      // Subtract one for the next barcode.
+      barcode -= 1;
+      // Pass the value and get a 14-digit barcode.
+      barcode = luhn.calculate(barcode);
```
EdwinGuzman, line 85: Example of this in the database:
Screen Shot 2020-05-11 at 11 18 31 AM

EdwinGuzman:
This is finally ready for review. The database race condition concerns should be all addressed now.

This now requires two local databases - one to run while the API is running and one that the tests used. The test database automatically creates and deletes the table after running tests, to ensure that it's properly inserting and selecting unique data. A test database is also set up in Travis and it looks like it's working. There's a flag when the tests run that tell it to use the database Travis sets up.
Once the real local database is created, the next time the app runs, a table and seed barcode data will be inserted. This might need to be updated when it's pushed to the QA and production environments.

db/index.js
```diff
@@ -2,65 +2,104 @@
...
+class BarcodesDb {
+  constructor(args) {
+    this.pool = new Pool({
+      user: args["user"],
+      host: args["host"],
+      database: args["database"],
+      password: args["password"],
+      port: args["port"],
+    });
+  }
```
leonardr, line 16: What's the difference between BarcodesDb here and BarcodeDb in app.js?

EdwinGuzman: This is the class. When this file is imported and invoked, an instance will be created and returned. The name in app.js is misleading and I can update that, but that `BarcodeDb` is the wrapper function that returns the instance of this class. I did it this way because I only wanted to have one instance of the class that connects to the database for every startup of the lambda. That way, multiple requests will use the same instance instead of creating a new instance for every request.

api/models/v0.3/modelBarcode.js
```diff
@@ -1,65 +1,241 @@
...
+   * @param {string} barcode
+   */
+  nextLuhnValidCode(barcode) {
+    if (!barcode || barcode.length !== 14) {
+      return;
     }
+
+    // Remove the last digit from the existing barcode since we need a
+    // 13-digit number to pass through the luhn-algorithm.
+    let newValidBarcode = Math.floor(parseInt(barcode, 10) / 10);
```
leonardr, line 51: It doesn't matter for our purposes, but if a barcode were to start with zeroes, this would strip the zeroes and we'd end up with a barcode smaller than 14 digits.

PR #53, opened 2020-05-15
Adding an endpoint to check if a patron is eligible to create dependent accounts.

EdwinGuzman:
## Description
Adds an endpoint and logic to check whether a patron is eligible to create dependent accounts.

## Motivation and Context

Resolves [DQ-304](https://jira.nypl.org/browse/DQ-304).

There's a new endpoint that whether or not a patron is eligible to create a dependent account.
Screen Shot 2020-05-14 at 1 17 23 PM
Screen Shot 2020-05-14 at 1 25 26 PM
The endpoint only requires a barcode. We first get the patron data object from the ILS with the `patronType` and `varFields` properties since those are the most important. 

There are 8 p-types that can create dependent accounts. If the patron has a `patronType` that is in the list (found in the code), then they can create dependent accounts! If not, they're not eligible. But, if the patron has the right p-type but has reached the limit of dependent accounts, they cannot create any more. The limit checking is done with the varFields object that looks like 
```
{ fieldTag: 'x', content: 'DEPENDENTS 1234,1235' }
```

There are many varFields objects but we only want the one with the `fieldTag` of "x" (there can be more than one). Then, look through the `content` string to find how many dependents they already have (there's more logic in the code).

## How Has This Been Tested?

Through Postman by hitting the endpoint. You can mock the returned patron data object in one of the functions but I created real patron accounts in the ILS with `patronTypes` of 1 and 2. I used their barcode to test end-to-end to get the different eligibility responses. Checking the `varFields` object with a `fieldTag` of "x" is hard to test right now because that part is another ticket.

## Checklist:

### DIFF
--- a/api/controllers/v0.3/DependentEligibilityAPI.js
+++ b/api/controllers/v0.3/DependentEligibilityAPI.js
@@ -1,0 +1,159 @@
+/* eslint-disable */
+const { NoILSClient, ILSIntegrationError } = require("../../helpers/errors");
+const IlsClient = require("./IlsClient");
+
+/**
+ *
+ */
+const DependentEligibilityAPI = (args) => {
+  const ilsClient = args["ilsClient"];
+
+  /**
+   * getPatronFromILS(barcode)
+   * This calls the ILS to get a patron data object from a barcode. Returns an
+   * error if the patron couldn't be found or there was an error with the ILS.
+   *
+   * @param {string} barcode
+   */
+  const getPatronFromILS = async (barcode) => {
+    if (!ilsClient) {
+      throw new NoILSClient(
+        "ILS Client not set in the Dependent Eligibility API."
+      );
+    }
+
+    try {
+      const response = await ilsClient.getPatronFromBarcodeOrUsername(barcode);
+
+      if (response.status !== 200) {
+        // The record wasn't found.
+        throw new Error("The patron couldn't be found.");
+      }
+
+      // Return the patron object.
+      return response.data;
+    } catch (error) {
+      throw new ILSIntegrationError(error.message);
+    }
+  };
+
+  /**
+   * checkPType(patronType)
+   * Checks if the input p-type is a valid p-type that can create
+   * dependent accounts in the ILS.
+   *
+   * @param {number} patronType
+   */
+  const checkPType = (patronType) =>
+    IlsClient.CAN_CREATE_DEPENDENTS.includes(patronType);
+
+  /**
+   * checkDependentLimit(varFields)
+   * This function assumes that the patron has a valid p-type, and so can
+   * create dependent accounts. Dependent accounts will be found in the
+   * varField array of a patron data object as an object with a fieldTag of "x".
+   * A varField object comes in the form of
+   *   { fieldtag: "", content: ""}
+   * This checks specificially if `content` has "DEPENDENTS" in the string. If
+   * it does, it then checks to see how many barcodes are in the string based
+   * on how many commas there are to separate the barcodes. If three barcodes
+   * are found, then the patron has reached their limit. Otherwise, they have
+   * one or two dependents and can create another. If the object with the
+   * fieldTag of "x" isn't found, or if it is found but it has other content,
+   * then we assume they don't have any dependents. They already have a
+   * valid p-type so let them create dependents.
+   *
+   * @param {array} varFields
+   */
+  const checkDependentLimit = (varFields) => {
+    // We only want varFields that have a fieldTag of "x".
+    const xFieldTags = varFields.filter((obj) => obj.fieldTag === "x");
+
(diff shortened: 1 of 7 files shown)


api/controllers/v0.3/IlsClient.js
```diff
@@ -297,34 +299,50 @@ IlsClient.PATRON_AGENCY_FIELD_TAG = "158";
...
+  DISABLED_METRO_NY_PTYPE: "Disabled Metro NY (3 Year)",
+  HOMEBOUND_NYC_PTYPE: "Homebound NYC (3 Year)",
+  SIMPLYE_METRO_PTYPE: "SimplyE Metro",
+  SIMPLYE_NON_METRO_PTYPE: "SimplyE Non-Metro",
   TEEN_METRO_PTYPE: "Teen Metro (3 Year)",
   TEEN_NYS_PTYPE: "Teen NY State (3 Year)",
   REJECTED_PTYPE: "Rejected",
   ILS_ERROR: "Unable to create in ILS",
 };
+IlsClient.CAN_CREATE_DEPENDENTS = [
```
EdwinGuzman, line 336: @killerpuppytails here's the list of the eight valid p-types that can create dependents. I set a temporary `101` p-type for the "Disabled Metro NY (3 Year)" and "Homebound NYC (3 Year)" patron types until we get the right code.

killerpuppytails: Thanks for tagging me @EdwinGuzman . This was very useful to look through so I ensure there's a path to accommodate teen parents, even while we're not supporting their p-type.

api/controllers/v0.3/DependentEligibilityAPI.js
```diff
@@ -0,0 +1,159 @@
...
+    let response = {};
+    const patron = await getPatronFromILS(barcode);
+    // First, check that they have a valid ptype to be able to create
+    // dependent accounts.
+    const hasValidPtype = checkPType(patron.patronType);
+
+    if (hasValidPtype) {
+      // Great, they have a valid ptype. Now check that they have not reached
+      // the dependent account limit.
+      const canCreateDependents = checkDependentLimit(patron.varFields);
```
adriana-alter, line 132: I realize this is not this branch's fault, but...why is there a limit?  What if someone has a lot of kids?

api/controllers/v0.3/DependentEligibilityAPI.js
```diff
@@ -0,0 +1,159 @@
...
+
+      if (!canCreateDependents) {
+        response["eligible"] = false;
+        response["description"] =
+          "This patron has reached the limit to create dependent patrons.";
```
adriana-alter, line 137: Shouldn't this be "to create dependent accounts" (like the next one is)?

EdwinGuzman: Oh, true.

api/controllers/v0.3/endpoints.js
```diff
@@ -351,7 +352,106 @@ async function createPatron(req, res) {
...
+  }
+
+  ilsClient =
+    ilsClient ||
+    IlsClient({
+      createUrl: process.env.ILS_CREATE_PATRON_URL,
+      findUrl: process.env.ILS_FIND_VALUE_URL,
+      ilsToken,
+      tokenUrl: process.env.ILS_CREATE_TOKEN_URL,
+      // ilsTokenTimestamp,
```
adriana-alter, line 431: Were these supposed to be commented out...?

api/controllers/v0.3/DependentEligibilityAPI.js
EdwinGuzman: That's the policy/requirement given.

EdwinGuzman: That scenario is definitely a possibility but there might be concerns we're not aware of that sets that limit.

api/controllers/v0.3/endpoints.js
EdwinGuzman: Yeah, I haven't refactored this part yet. I want to try passing in these values or setting them in the `IlsClient` itself, but since those are variables that the whole app cares about, it seems to be better to leave them at the app scope rather than in the class scope. It's also related to how the lambda startups and runs.

api/controllers/v0.3/DependentEligibilityAPI.js
killerpuppytails: I can speak to the limitation requirements via Slack, if you're interested.

EdwinGuzman: Yes, please.

adriana-alter: Yes--I'd love to hear about that too!

PR #54, opened 2020-05-19
Link dependent accounts in the ILS

EdwinGuzman:
## Description

Updated `DependentEligibilityAPI` to `DependentAccountAPI` since it does more than just check for a patron's eligibility to create dependent juvenile accounts. Now, we need to "link" newly created dependent account with its parent account, and vice-versa.

## Motivation and Context

This resolves [DQ-305](https://jira.nypl.org/browse/DQ-305) and [DQ-306](https://jira.nypl.org/browse/DQ-306). There's a bit of code overlap that can be refactored in `DependentEligibilityAPI` but I want to get this out on QA by tomorrow so Malcolm and Ettore can test it, so that'll be a different ticket. This makes use of the `varFields` property to store values that will be used to link a patron to its dependent and vice-versa. The object this is stored in looks like: `{ fieldTag: 'x', content: 'DEPENDENT ...'}`.

## How Has This Been Tested?

Using the ILS and Postman.
1) First, create a patron using the `simplye` policy. The ILS will return the patron ID.
2) We can use that ID to find the barcode using the ILS's Swagger doc.
3) Now we can hit `/patrons/dependent-eligibility` with the patron's barcode to find if they are eligible to create a dependent juvenile card.
4) The other endpoint is `/patrons/dependents` and that can be hit with the patron's barcode along with a username, pin, name, and birthdate (some fields can change soon). Two events happen here:

  - First, the dependent juvenile card is created with the request values of username, pin, name, and birthdate. The address and email come from the parent. This card has a ptype of `simplyeJuvenile` code 4. A barcode is generated for this new card. To link this card to the parents, we tell the ILS that it is a "dependent of {parent's barcode}". For a parent barcode of "123456789", this is sent in the `varFields` as `varFields: [{...}, { fieldTag: 'x', content: 'DEPENDENT OF 123456789' }, {...}]`.
  - The juvenile card is complete at this point. Now the parent card needs to be updated in a PUT request to the ILS. All this does is update a patron's data with the same `varFields` `fieldTag` but now we use the dependent's barcode in the `content` property. For a dependent juvenile barcode of "987654321", the `varFields` value looks like `varFields: [{...}, { fieldTag: 'x', content: 'DEPENDENTS 987654321' }, {...}]`. Every new dependent is added on, so the third dependent will look like `{ fieldTag: 'x', content: 'DEPENDENTS 987654321,987654322,987654323' }`.

5)The ILS returns a 204 "no body content" for a successful patron update. So the response we return is the response from creating the dependent juvenile card, and just a hardcoded "okay" to denote that the parent patron was updated. If it wasn't updated, an error would be thrown and we'd catch that.

## Checklist:

### DIFF
--- a/api/controllers/v0.3/DependentAccountAPI.js
+++ b/api/controllers/v0.3/DependentAccountAPI.js
@@ -2,8 +2,14 @@
 const { NoILSClient, ILSIntegrationError } = require("../../helpers/errors");
 const IlsClient = require("./IlsClient");
 
+// A parent patron is only allowed to create three dependent juvenile accounts.
+const DEPENDENT_LIMIT = 3;
+
 /**
- *
+ * DependentAccountAPI
+ * A class with the main purpose to check a patron's eligibility to create
+ * dependent juvenile cards and to update a patron account with the information
+ * of the dependent juvenile cards.
  */
 const DependentAccountAPI = (args) => {
   const ilsClient = args["ilsClient"];
@@ -30,15 +36,15 @@ const DependentAccountAPI = (args) => {
     // Set the fetched patron data object into the global variable so
     // it can be accessed by `getPatron`. This is specifically set here and not
     // in `getPatronFromILS` because we want to make sure that the eligibility
-    // check was ran in order to retrieve a patron.
+    // check was run in order to retrieve a patron.
     parentPatronData = patron;
-    // First, check that they have a valid ptype to be able to create
-    // dependent accounts.
-    const hasValidPtype = checkPType(patron.patronType);
+    // First, check that they have an eligible ptype that allows them to
+    // create dependent accounts.
+    const hasEligiblePtype = checkPType(patron.patronType);
 
-    if (hasValidPtype) {
-      // Great, they have a valid ptype. Now check that they have not reached
-      // the dependent account limit.
+    if (hasEligiblePtype) {
+      // Great, they have an eligible ptype. Now check that they have not
+      // reached the dependent account limit.
       const canCreateDependents = checkDependentLimit(patron.varFields);
 
       if (!canCreateDependents) {
@@ -88,7 +94,7 @@ const DependentAccountAPI = (args) => {
 
   /**
    * checkPType(patronType)
-   * Checks if the input p-type is a valid p-type that can create
+   * Checks if the input p-type is an eligible p-type that can create
    * dependent accounts in the ILS.
    *
    * @param {number} patronType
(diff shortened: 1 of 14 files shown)


api/controllers/v0.3/DependentAccountAPI.js
```diff
@@ -5,8 +5,57 @@ const IlsClient = require("./IlsClient");
...
+        "ILS Client not set in the Dependent Eligibility API."
+      );
+    }
+
+    let response = {};
+    const patron = await getPatronFromILS(barcode);
+    // Set the fetched patron data object into the global variable so
+    // it can be accessed by `getPatron`. This is specifically set here and not
+    // in `getPatronFromILS` because we want to make sure that the eligibility
+    // check was ran in order to retrieve a patron.
```
adriana-alter, line 33: Little typo--"was RUN"

api/controllers/v0.3/DependentAccountAPI.js
```diff
@@ -5,8 +5,57 @@ const IlsClient = require("./IlsClient");
...
+    parentPatronData = patron;
+    // First, check that they have a valid ptype to be able to create
+    // dependent accounts.
+    const hasValidPtype = checkPType(patron.patronType);
+
+    if (hasValidPtype) {
+      // Great, they have a valid ptype. Now check that they have not reached
+      // the dependent account limit.
+      const canCreateDependents = checkDependentLimit(patron.varFields);
```
adriana-alter, line 42: It might be helpful to specify in the comment what the limit is (unless we think it's about to change)

api/controllers/v0.3/DependentAccountAPI.js
```diff
@@ -5,8 +5,57 @@ const IlsClient = require("./IlsClient");
...
+
+      if (!canCreateDependents) {
+        response["eligible"] = false;
```
adriana-alter, line 45: It's confusing that it's called "eligible" here but "valid" in the comments and variable names; took me a moment to realize this was the same thing...

api/controllers/v0.3/DependentAccountAPI.js
EdwinGuzman: Yeah, I wrote that the limit is three in the specific function that does that, but can include it here.

api/controllers/v0.3/DependentAccountAPI.js
EdwinGuzman: It's using "valid" for the ptype but not for eligibility though. I can clean it up.

api/controllers/v0.3/DependentAccountAPI.js
```diff
@@ -105,55 +154,123 @@ const DependentEligibilityAPI = (args) => {
...
+      // The record wasn't found and couldn't be updated.
+      throw new Error("The parent patron couldn't be updated.");
     }
 
     return response;
   };
 
+  /**
+   * formatDependentAddress(address)
+   * A dependent account has the address as its parent account. The address
```
adriana-alter, line 234: Do you mean "has the SAME address as its parent account"?

EdwinGuzman: Yep.

api/controllers/v0.3/IlsClient.js
```diff
@@ -317,6 +401,9 @@ IlsClient.ADULT_METRO_PTYPE = 10;
 IlsClient.ADULT_NYS_PTYPE = 11;
 IlsClient.SENIOR_METRO_PTYPE = 20;
 IlsClient.SENIOR_NYS_PTYPE = 21;
+IlsClient.SIMPLYE_JUVENILE = 4;
```
adriana-alter, line 404: What's the difference between "juvenile" and "juvenile only"?

EdwinGuzman: I'm not sure. @killerpuppytails ?

killerpuppytails: Juvenile means all access - it doesn't have any restrictions for checkout. The reason we have it is to allow appropriate database service access. Juvenile only does have restrictions for checkout.

adriana-alter:
This is good--just a couple of minor questions and typos.

PR #58, opened 2020-05-28
Return better values

EdwinGuzman:
## Description

This adds better correct and error responses from the API.

## Motivation and Context

This resolves [DQ-308](https://jira.nypl.org/browse/DQ-308). There were feedback requests from the API now that it's on the QA server.
1) Increase barcode values rather than decrease - This was done in #56 and was big enough to require its own ticket.

This ticket/PR resolves the rest of the feedback:
2) Return better error values - after chatting with @ettore . Also removing all `new Error` in favor of specific types of errors which will help when logging them in the near future.
3) Add copy from @killerpuppytails
4) Return the newly created dependent juvenile account's barcode (along with other values)

### `400`s for the following:
* if the parent account is not eligible to create dependent accounts if they reached the limit.
Screen Shot 2020-05-28 at 10 27 32 AM
Screen Shot 2020-05-28 at 11 03 11 AM

* if the account does not have the correct p-type.
Screen Shot 2020-05-28 at 10 59 17 AM

* if the username is invalid
Screen Shot 2020-05-28 at 11 12 10 AM

* if the username is unavailable
Screen Shot 2020-05-28 at 11 12 24 AM

* incorrect pin
Screen Shot 2020-05-28 at 10 18 33 AM

* missing a required field
Screen Shot 2020-05-28 at 10 14 49 AM

### `200`s for the following:

* `/dependent-eligibility` - patron can create dependent juvenile cards
Screen Shot 2020-05-28 at 11 04 48 AM

* `/dependent` - returning more details of the dependent and the parent accounts
Screen Shot 2020-05-28 at 10 19 38 AM

### swagger doc 
Similar examples can be found in the swagger doc. Unfortunately, Swagger does not support multiple examples for one error code so there's only one per each. Another reason why I added more examples in this PR.

Screen Shot 2020-05-28 at 10 28 55 AM
Screen Shot 2020-05-27 at 5 28 20 PM

## How Has This Been Tested?

Locally through Postman and through the Swagger documentation.

## Checklist:

If I'm missing anything, let me know @ettore .

### DIFF
--- a/api/controllers/v0.3/DependentAccountAPI.js
+++ b/api/controllers/v0.3/DependentAccountAPI.js
@@ -1,0 +1,282 @@
+/* eslint-disable */
+const { NoILSClient, ILSIntegrationError } = require("../../helpers/errors");
+const IlsClient = require("./IlsClient");
+
+// A parent patron is only allowed to create three dependent juvenile accounts.
+const DEPENDENT_LIMIT = 3;
+
+/**
+ * DependentAccountAPI
+ * A class with the main purpose to check a patron's eligibility to create
+ * dependent juvenile cards and to update a patron account with the information
+ * of the dependent juvenile cards.
+ */
+const DependentAccountAPI = (args) => {
+  const ilsClient = args["ilsClient"];
+  let parentPatronData;
+
+  /**
+   * isPatronEligible(barcode)
+   * The main function of this class. It gets a patron data object from the ILS
+   * and it verifies it has the correct p-type. If it does, then it checks if
+   * it can create another dependent account. It returns a response stating
+   * whether the patron is eligible or not and a description if they are not.
+   *
+   * @param {string} barcode
+   */
+  const isPatronEligible = async (barcode) => {
+    if (!ilsClient) {
+      throw new NoILSClient(
+        "ILS Client not set in the Dependent Eligibility API."
+      );
+    }
+
+    let response = {};
+    const patron = await getPatronFromILS(barcode);
+    // Set the fetched patron data object into the global variable so
+    // it can be accessed by `getPatron`. This is specifically set here and not
+    // in `getPatronFromILS` because we want to make sure that the eligibility
+    // check was run in order to retrieve a patron.
+    parentPatronData = patron;
+    // First, check that they have an eligible ptype that allows them to
+    // create dependent accounts.
+    const hasEligiblePtype = checkPType(patron.patronType);
+
+    if (hasEligiblePtype) {
+      // Great, they have an eligible ptype. Now check that they have not
+      // reached the dependent account limit.
+      const canCreateDependents = checkDependentLimit(patron.varFields);
+
+      if (!canCreateDependents) {
+        response["eligible"] = false;
+        response["description"] =
+          "This patron has reached the limit to create dependent accounts.";
+      } else {
+        response["eligible"] = true;
+        response["description"] = "This patron can create dependent accounts.";
+      }
+    } else {
+      response["eligible"] = false;
+      response["description"] = "This patron does not have an eligible ptype.";
+    }
+
+    return response;
+  };
+
+  /**
+   * getPatronFromILS(barcode)
+   * This calls the ILS to get a patron data object from a barcode. Returns an
+   * error if the patron couldn't be found or there was an error with the ILS.
+   *
+   * @param {string} barcode
+   */
+  const getPatronFromILS = async (barcode) => {
+    if (!ilsClient) {
+      throw new NoILSClient(
+        "ILS Client not set in the Dependent Eligibility API."
+      );
+    }
+
+    try {
+      const response = await ilsClient.getPatronFromBarcodeOrUsername(barcode);
+
+      if (response.status !== 200) {
+        // The record wasn't found.
+        throw new Error("The patron couldn't be found.");
+      }
+
+      // Return the patron object.
+      return response.data;
+    } catch (error) {
+      throw new ILSIntegrationError(error.message);
+    }
+  };
+
+  /**
+   * checkPType(patronType)
+   * Checks if the input p-type is an eligible p-type that can create
+   * dependent accounts in the ILS.
+   *
+   * @param {number} patronType
+   */
+  const checkPType = (patronType) =>
+    IlsClient.CAN_CREATE_DEPENDENTS.includes(patronType);
+
+  /**
+   * checkDependentLimit(varFields)
+   * This function assumes that the patron has an eligible p-type, and so can
+   * create dependent accounts. Dependent accounts will be found in the
+   * varField array of a patron data object as an object with a fieldTag of "x".
+   * A varField object comes in the form of
+   *   { fieldtag: "", content: ""}
+   * This checks specificially if `content` has "DEPENDENTS" in the string. If
+   * it does, it then checks to see how many barcodes are in the string based
+   * on how many commas there are to separate the barcodes. If the limit of
+   * barcodes are found (DEPENDENT_LIMIT), then the patron has reached their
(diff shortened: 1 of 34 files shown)


api/controllers/v0.3/DependentAccountAPI.js
```diff
@@ -32,7 +40,10 @@ const DependentAccountAPI = (args) => {
     }
 
     if (!barcode) {
-      throw new Error("Invalid Request: No barcode passed");
+      throw new InvalidRequest("No barcode passed.");
+    }
+    if (barcode.length !== 14) {
+      throw new InvalidRequest("The barcode passed is not 14-digits.");
```
adriana-alter, line 46: Why is this hyphenated?

api/controllers/v0.3/DependentAccountAPI.js
```diff
@@ -170,7 +205,41 @@ const DependentAccountAPI = (args) => {
...
+      return;
+    }
+
+    const varFields = parentPatronData.varFields || [];
+    let dependents;
+    const xFieldTags = varFields.filter((obj) => obj.fieldTag === "x");
+    // No varFields were found, so we can assume the patron doesn't
+    // have any dependent accounts yet.
+    if (xFieldTags.length === 0) {
+      dependents = "DEPENDENTS ";
```
adriana-alter, line 219: Is the space at the end of the string supposed to be there?

api/controllers/v0.3/DependentAccountAPI.js
```diff
@@ -170,7 +205,41 @@ const DependentAccountAPI = (args) => {
...
+      // "DEPENDENTS x,x,x"
+      // where `x` is a barcode. First get the varField that has "DEPENDENTS".
+      const dependentsVarField = xFieldTags.find(
+        (obj) => obj.content.indexOf("DEPENDENTS") !== -1
+      );
+
+      // There are varFields with a fieldTag of "x" but none with "DEPENDENTS".
+      // We can assume the patron doesn't have any dependents already.
+      if (!dependentsVarField) {
+        dependents = "DEPENDENTS ";
```
adriana-alter, line 231: Maybe-extra space again

api/controllers/v0.3/endpoints.js
```diff
@@ -446,12 +449,18 @@ async function checkDependentEligibility(req, res) {
     status = 200;
   } catch (error) {
     response = modelResponse.errorResponseData(
-      collectErrorResponseData(error.status || 400, "", error.message, "", "") // eslint-disable-line comma-dangle
+      collectErrorResponseData(
+        error.status || 400,
+        error.type || "",
+        error.message,
+        "",
```
adriana-alter, line 456: What are the empty strings for?

api/helpers/errors.js
```diff
@@ -44,10 +46,87 @@ class ILSIntegrationError extends Error {
...
+    this.status = 400;
+  }
+}
+
+class ExpiredAccount extends Error {
+  constructor() {
+    super();
+    this.type = "expired-account";
+    this.name = "ExpiredAccount";
+    this.message = "Your card has expired. Please try applying again.";
```
adriana-alter, line 104: This phrasing seems potentially confusing--won't people think it means they should just retry the thing they just did that led to this error?

api/controllers/v0.3/DependentAccountAPI.js
EdwinGuzman: True 👀

api/controllers/v0.3/DependentAccountAPI.js
EdwinGuzman: It is. This is only for the response back to the requester so it's easy to add the newly created dependent juvenile barcode to this string. It's in `/api/controllers/v0.3/endpoints.js` L682 - https://github.com/NYPL/dgx-patron-creator-service/pull/58/files#diff-04f337212864a7293b4938d02be88954R682

api/helpers/errors.js
EdwinGuzman: That's true. @killerpuppytails , I got this from the document. Should it be updated or should the clients handle this error?

api/controllers/v0.3/endpoints.js
EdwinGuzman: This is from the `collectErrorResponseData` function and the last two parameters are for `title` and `debugMessage`. For this specific example, the error object will be something like:
```
{
    "status": 400,
    "type": "unavailable-username",
    "message": "This username is unavailable. Please try another.",
}
```

and `title` and `debugMessage` aren't necessary. It also adds a `detail` key which isn't used. The most important things are returned right now and I'll remove the rest in another pass. It was already in the code but I don't want to delete it now in case it breaks any previous endpoints or existing functions.

api/controllers/v0.3/DependentAccountAPI.js
```diff
@@ -52,21 +71,34 @@ const DependentAccountAPI = (args) => {
       const canCreateDependents = checkDependentLimit(patron.varFields);
 
       if (!canCreateDependents) {
-        response["eligible"] = false;
-        response["description"] =
-          "This patron has reached the limit to create dependent accounts.";
+        throw new NotEligibleCard(
+          "You have reached the limit of dependent cards you can receive via online application."
```
ettore, line 75: for this business logic error and the ones below related to `NotEligibleCard`, are you still returning a response with `eligible` and `description` fields as we had before?

ettore: and do these business logic errors still end up with a 200 status code?

EdwinGuzman: No, those errors are now being returned as a 400 response. So now you'll get a response like:

```
{
    status: 400,
    type: 'not-eligible-card',
    message: '....'
}
```

Check this PR's description for more examples.

api/helpers/errors.js
killerpuppytails: This is actually the same language that would be used by Ask. Therefore, no update. If we need to, the clients can include custom error text.

adriana-alter: In that case, I think this is all set!

api/controllers/v0.3/DependentAccountAPI.js
```diff
@@ -52,21 +71,34 @@ const DependentAccountAPI = (args) => {
...
+        );
       } else {
         response["eligible"] = true;
```
ettore, line 78: Is there ever a situation where `response["eligible"] == false` ?  I have logic in the client that handles that, but if this is impossible I can simplify and remove.

EdwinGuzman: There won't be a situation like that anymore. If the patron is not eligible, an error will be thrown instead of returning a response object that includes `eligible: false`. I think it makes it easier if you call the API and get a 400 with bad data, rather than getting a 200 and having to parse it to see what the eligibility is.

ettore: that makes sense to me too. In this case then I can remove parsing of the successful response which essentially is like a 204 for me

api/helpers/errors.js
```diff
@@ -44,10 +46,87 @@ class ILSIntegrationError extends Error {
   }
 }
 
+class PatronNotFound extends Error {
+  constructor() {
+    super();
+    this.type = "patron-not-found";
+    this.name = "PatronNotFound";
+    this.message = "The patron couldn't be found.";
```
ettore, line 54: Can this message be more user-friendly? It is kind of confusing when displayed on the client.

EdwinGuzman: Sure. I'm going to update it to `"The patron couldn't be found in the ILS with the barcode or username."` since this happens when requesting a patron from the ILS.

PR #62, opened 2020-05-29
Updated logs

EdwinGuzman:
## Description

This updates the Winston package and updates the logging file accordingly. Now, it's also logging to a local file (which is ignored by git locally) but useful if we want to keep it on AWS.

## Motivation and Context

No ticket, just improvement based on work that @gkallenberg did recently.

## How Has This Been Tested?

Locally:

Screen Shot 2020-05-29 at 11 19 04 AM

The file `/log/dgx-patron-creator-service.log` will log similar messages:
```
{"timestamp":"2020-05-29T15:18:47.584Z","levelCode":3,"level":"ERROR","message":"database table barcodes already exists, continuing","appTag":"dgx-patron-creator-service","pid":"80565"}
{"timestamp":"2020-05-29T15:18:47.596Z","levelCode":3,"level":"ERROR","message":"barcodes table already has the initial value","appTag":"dgx-patron-creator-service","pid":"80565"}
{"timestamp":"2020-05-29T15:18:51.778Z","levelCode":3,"level":"ERROR","message":"status_code: 400, type: invalid-request, message: The barcode passed is not 14 digits., response: ","appTag":"dgx-patron-creator-service","pid":"80565"}
```

## Checklist:


api/helpers/Logger.js
```diff
@@ -1,9 +1,12 @@
 const winston = require('winston');
```
EdwinGuzman, line 1: @gkallenberg thanks for the idea of updating this file. Hopefully, we can create a general pattern or package each app can import rather than copying and pasting this file.

PR #63, opened 2020-05-29
Update npm packages

EdwinGuzman:
## Description

Updates packages to fix any security issues through `npm audit`. I removed the swagger-related packages because they are not even used in this repo. I can see why they were added there but they were never implemented. Those packages are outdated and gave almost all of the security warnings. For posterity, I use http://editor.swagger.io/ to edit and update the current swagger json and yaml files.

Current audit on `master`:
Screen Shot 2020-05-29 at 3 53 47 PM

Current audit on this branch:
Screen Shot 2020-05-29 at 3 45 41 PM

## Motivation and Context

Resolves [DQ-314](https://jira.nypl.org/browse/DQ-314). No one likes security issues.

## How Has This Been Tested?

Locally ran tests and used postman to verify everything is still working as its suppose to.

## Checklist:


adriana-alter:
Does this necessitate a different start-up process?  When I try to run it, I get this in the terminal:
Screen Shot 2020-06-01 at 2 08 42 PM
And this in the browser:
Screen Shot 2020-06-01 at 2 17 22 PM

EdwinGuzman:
No. That's working fine as is. That error message is a database error but not an error what would crash the app so it's okay to see, and you'll see it on every start once the database is initialized.
You're not meant to view the API in the browser so that's a valid response you see. You can curl or use Postman to hit the endpoints, since they are also authenticated endpoints.

adriana-alter:
In that case, this looks good!

PR #70, opened 2020-06-12
Service Objects address validation implementation

EdwinGuzman:
## Description

This implements the Service Objects API into a client to use to validate addresses. Currently, this PR sets up the `/api/v0.3/validation/address` endpoint. The address validation is not hooked up to patron creation just yet. That's for another PR.

Note: The code is ready to review but there may be business logic that needs to be updated if @killerpuppytails finds the responses in the following section wrong.

## Motivation and Context

This resolves [DQ-316](https://jira.nypl.org/browse/DQ-316). The `/validate/address` endpoint takes an `address` and `policyType` (defaults to "simplye") to validate an address and also get the correct type of card.

### A couple of errors:
1) If the license key is wrong or something happened, the call errors. In the old Card Creator it use to continue and just return a temporary card but maybe we don't want this anymore (@killerpuppytails, does this sound right?)
Screen Shot 2020-06-12 at 12 45 38 PM

2) whoops, didn't enter a correct street address:
Screen Shot 2020-06-12 at 12 50 30 PM

### Responses:
1) Someone outside NY tried to apply for a card:
Screen Shot 2020-06-12 at 12 35 28 PM

The response returns no card type since they are not in New York. Currently, the `simplye` policy returns this response. If the `webApplicant` policy is used, a different response is returned that tells the user they are getting a temporary card, even though they are not in NY. @killerpuppytails, does this sound right? Or should everyone outside NY regardless of policy not get a card?:
Screen Shot 2020-06-12 at 12 35 39 PM

2) Similar to the above, if an address in NY is entered, the response is different based on the policy.
Here is an address with "simplye" policy and the card is temporary:
Screen Shot 2020-06-12 at 12 37 07 PM

and here is the same address with "webApplicant" policy and the card is a standard card:
Screen Shot 2020-06-12 at 12 36 59 PM

## How Has This Been Tested?

Locally through Postman.

## Checklist:

### DIFF
--- a/api/controllers/v0.3/AddressValidationAPI.js
+++ b/api/controllers/v0.3/AddressValidationAPI.js
@@ -3,7 +3,7 @@ const ServiceObjectsClient = require("./ServiceObjectsClient");
 const {
   SOAuthorizationError,
   SODomainSpecificError,
-  SOIntegrationError,
+  SONoLicenseKeyError,
 } = require("../../helpers/errors");
 const Address = require("../../models/v0.3/modelAddress");
 const { Card } = require("../../models/v0.3/modelCard");
@@ -12,13 +12,8 @@ const Policy = require("../../models/v0.3/modelPolicy");
 /**
  * A class that uses Service Objects to validate addresses.
  */
-const AddressValidationAPI = (args) => {
+const AddressValidationAPI = (args = {}) => {
   const soLicenseKey = args["soLicenseKey"];
-  if (!soLicenseKey) {
-    throw new SOIntegrationError(
-      "No credentials for Service Objects were passed in AddressValidationAPI."
-    );
-  }
 
   const ALTERNATE_ADDRESSES_TYPE = "alternate-addresses";
   const UNRECOGNIZED_ADDRESS_TYPE = "unrecognized-address";
@@ -40,7 +35,20 @@ const AddressValidationAPI = (args) => {
     },
   };
 
+  /**
+   * validate(address, policyType)
+   * Calls the Service Objects client to validate an address and returns a
+   * response or throws an error.
+   * @param {Object} address
+   * @param {string} policyType
+   */
   const validate = async (address, policyType = "simplye") => {
+    if (!soLicenseKey) {
+      throw new SONoLicenseKeyError(
+        "No credentials for Service Objects were passed in AddressValidationAPI."
+      );
+    }
+
     let response;
     try {
       const addresses = await validateAddress(address);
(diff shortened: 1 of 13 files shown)


killerpuppytails:
So this is tricky. The SO error will probably require two conditions. If by whatever function it is determined that the applicant IS in New York State and yet the Service Objects call returns an error, yes absolutely we give them a temporary card, but we tell them why and ask them to try again in a certain number of days. (I'll get copy for you for that.) If the function to determine location is absent or we can tell the applicant is not in New York State, they should receive an error saying it does not look like they're in New York and asking them to contact gethelp at nypl dot org if they believe this is in error.

For "home address not in New York" there should be a condition that allows them to enter a work address, which can then be validated, and if that can't be validated, again they should receive an error saying it does not look like they're in New York and asking them to contact gethelp at nypl dot org if they believe this is in error.

I think I've covered both your questions, but I also think there's another condition I'm not thinking of, so I might comment again :)

EdwinGuzman:
Yeah, there are a couple of conditions and some of the existing logic was confusing to read through. There was also always the option for allowing work addresses but it was never used and none of clients allowed that, so I'm not sure how well that logic is.

The first part makes sense so I'll rewrite that with updated copy once it's set, thanks.

EdwinGuzman:
I'll add the work address logic in a different PR since that's a slightly different topic. The logic is correct, however, that a "simplye" applicant (currently through SimplyE) will always get a temporary card and a "web applicant" (currently through the web app) will always get a standard card, right?

killerpuppytails:
OK just taking it a step back to confirm: If an applicant comes through SimplyE and we detect they are in NYS but the SO call fails, they receive a temp card. If an applicant comes through SimplyE, we detect they are in NYS, the SO call succeeds and the address is validated, they should receive a 3-yr SimplyE card, not a temp one.

I'll verify if this same logic should apply to the web applicant, they might need to get a temp card before they are issued a standard one or we might issue them a digital-only card as well, that they can upgrade at a branch.

EdwinGuzman:
Does the following make sense? I'm not sure what the difference between "simplye" and "webApplicant" would be in this case but I know I'm missing something. This doesn't include the "work address" input, just the home address.

simplye:
1) outside NYS, SO failed                                -> no card
2) outside NYS, SO succeeded                      -> no card
2) in NYS, SO failed                                         -> temp card
4) in NYS, SO succeeded, invalid address    -> temp card
3) in NYS, SO succeeded, valid address       -> standard

webApplicant:
1) outside NYS, SO failed                                -> no card
2) outside NYS, SO succeeded                      -> no card
2) in NYS, SO failed                                         -> temp card
4) in NYS, SO succeeded, invalid address    -> temp card
3) in NYS, SO succeeded, valid address       -> standard

killerpuppytails:
yes those logics are correct! I will validate with Rosa what p-types will be supported by the nypl.org entry point vs the SimplyE entry point.

EdwinGuzman:
Thanks, Risa.

@adriana-alter , the last commit is a big one and makes some big changes. Instead of returning an error as a response from the endpoint, the error is caught and passed to the next step. In this case, if Service Objects throws any type of error, it is caught and passed to the next step where the "unconfirmed" (since it can't be validated) address is checked naively if it's in NYS. If it is, a temporary card is returned. More of the logic is discussed in the previous comments.

Here are some screenshots from Postman:
1) I mocked a "city not found" error from service objects for an address in NYS and an address outside of NYS. They are both "unrecognized-addresses" (since SO returned an error) but the in-NYS address returns a "cardType" of "temporary".
Screen Shot 2020-06-16 at 12 12 39 PM
Screen Shot 2020-06-16 at 12 12 50 PM

2) Service Objects works and returns no card type, "temporary", or "standard" (the library address returns temporary because it's not residential). All addresses are "valid-address" because SO was able to validate them, but it doesn't mean the policy allows for a card so that's checked after.

Screen Shot 2020-06-16 at 12 13 31 PM
Screen Shot 2020-06-16 at 12 20 13 PM
Screen Shot 2020-06-16 at 12 13 43 PM

adriana-alter:
@EdwinGuzman The changes look good!

PR #71, opened 2020-06-18
Card Creation with Address Validation

EdwinGuzman:
## Description

Makes `/api/v0.3/createPatron` validate addresses by connecting all the pieces together. The endpoint now also returns more values.

Example input:
```
{
  username: "username",
  name: "name",
  address: {
    line1: "...",
    city: "...",
    state: "...",
    zip: "...",
  },
  pin: "1234",
  birthdate: "01-01-1988",
  email: "email@email.com"
}
```

Example response:
```
{
  status: 200,
  link: "link/to/patron/in/ils",
  type: "card-granted",
  barcode: "12345678912345",
  username: "username",
  pin: "1234",
  temporary: false,
  message: "Your library card has been created."
}
```

This updates the previous PR and better separates concerns of what each class does. So now, `AddressValidationAPI` only handles address validation and returns a response based on a correct or error response from Service Objects. Before it also used to check the address against a card's policy type to find out if it was a denied, temporary, or standard card. That is now the job of the `Card` class. Another reason for doing this is that the `Address` class depended on `AddressValidationAPI` (which makes sense) but I was trying to create new `Address` classes in `AddressValidationAPI` and this caused a circular dependency and nightmares.

Errors in the `/createPatron` get returned just like before, for invalid or unavailable username, or for missing values (pin, name, etc).

`/api/v0.3/validations/address` works just as before even though that has changed in `endpoints.js`.

There's also some cleanup and refactoring. There's more I want to do but can wait for another PR, for example, passing an `AddressValidationAPI` client rather than passing the `soLicenseKey` every time a new `Address` class is created (which is often).

## Motivation and Context

This resolves [DQ-320](https://jira.nypl.org/browse/DQ-320).

## How Has This Been Tested?

Locally through Postman:

#### Errors:
Not passing a complete address:
Screen Shot 2020-06-18 at 10 59 27 AM

Username is unavailable:
Screen Shot 2020-06-18 at 10 59 10 AM

#### Types of cards:
Denied because you're not in NYS!! 😞 
Screen Shot 2020-06-18 at 10 58 59 AM

Temporary because the address is not residential or it couldn't be fully validated or is unrecognized by SO 😕 
Screen Shot 2020-06-18 at 10 58 42 AM

Standard! 😄 
Screen Shot 2020-06-18 at 10 58 21 AM

## Checklist:

### DIFF
--- a/api/models/v0.3/modelAddress.js
+++ b/api/models/v0.3/modelAddress.js
@@ -150,9 +150,8 @@ class Address {
    * validate(isWorkAddress)
    * Simple validation to make sure the address length is the proper length. If
    * it is, it then validates the address in Service Objects.
-   * @param {boolean} isWorkAddress
    */
-  async validate(isWorkAddress = false, policyType = "simplye") {
+  async validate() {
     const fullAddressLength = (this.address.line1 + this.address.line2).length;
     if (fullAddressLength > 100) {
       this.errors["line1"] =
--- a/api/models/v0.3/modelAddress.js
+++ b/api/models/v0.3/modelAddress.js
@@ -136,15 +136,14 @@ class Address {
   }
 
   /**
-   * validateInAPI(isWorkAddress)
-   * @param {boolean} isWorkAddress
+   * validateInAPI()
    */
-  async validateInAPI(isWorkAddress = false, policyType = "simplye") {
+  async validateInAPI() {
     const { validate } = AddressValidationAPI({
       soLicenseKey: this.soLicenseKey,
     });
-    let response = await validate(this.address, isWorkAddress, policyType);
-    return response;
+
+    return await validate(this.address);
   }
 
   /**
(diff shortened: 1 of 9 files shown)


api/models/v0.3/modelAddress.js
```diff
@@ -151,44 +136,39 @@ class Address {
...
+   * validate(isWorkAddress)
+   * Simple validation to make sure the address length is the proper length. If
+   * it is, it then validates the address in Service Objects.
    */
-  validatedVersion(isWorkAddress = undefined) {
+  async validate() {
+    const fullAddressLength = (this.address.line1 + this.address.line2).length;
+    if (fullAddressLength > 100) {
+      this.errors["line1"] =
+        "Address lines must be less than 100 characters combined.";
```
adriana-alter, line 158: Can we tell the user how many characters they're currently at?  (Also, aren't there any legitimate cases in which someone's address is over 100 characters...?)

adriana-alter:
The code looks good--just curious about that one error message!

api/models/v0.3/modelAddress.js
EdwinGuzman: Yeah, it's possible to return how many characters they are over. There _could_ be cases there the length is per 100 but I'm not sure. This is an error that's carried over from the previous code. I think it makes some sense but I can't think of outliers right now.

PR #72, opened 2020-06-30
Work address

EdwinGuzman:
## Description

This introduces an optional `workAddress` parameter to both the `/api/v0.3/validations/address` and `/api/v0.3/patrons` endpoints. 

1. `/api/v0.3/validations/address` - This is used just to validate an address directly through Service Objects without worrying about other values. `isWorkAddress` is optional and defaults to `false` and `policyType` is optional and defaults to `simplye`. This endpoint returns whether the input address is valid or invalid, and what type of card can be created with that address. 
Return example values:

| type | message | cardType | status | address | originalAddress | reason | error |
| --- | --- | --- | --- | --- | --- | --- | --- |
| "valid-address" | "The library card will be a standard library card." | "standard" | 200 | updated address from SO | original address input | - | - |
| "valid-address" | "The library card will be a temporary library card." | "temporary" | 200 | updated address from SO | original address input | "The home address is in NYC but is not residential" | - |
| "valid-address" | "Library cards are only available for residents of New York State or students and commuters working in New York City." | null" | 200 | updated address from SO | original address input | - | - |
| "unrecognized-address" | "Library cards are only available for residents of New York State or students and commuters working in New York City." | null | 400 | - | original address input |  - | `"error": { ..., "message": "Please input either zip code or both city and state." }`|

An example input:
```
{
	"address": {
		"line1": "24 Kilmer Rd #357",
		"line2": "",
		"city": "Edison",
		"state": "NJ",
		"zip": "00817"
	},
	"isWorkAddress": false,
	"policyType": "simplye"
}
```

Screen Shot 2020-07-01 at 1 33 18 PM

1. `/api/v0.3/patrons` - This endpoint now accepts a `workAddress`. The `workAddress` will mostly be used and checked when a patron is outside of NYS but works in NYC. If that's the case, they get a temporary card. 

@killerpuppytails ,  the screenshot below is probably the main use case and response when creating a new patron with an address outside NYS but a work address in NYC. I'll create a document with test values and responses I've been using to test the logic out and I'll eventually attach it to this PR.

Return example values:

| status | link | type | barcode | username | pin | temporary | message | patronId |
| --- | --- | --- | --- | --- | --- | --- | --- | --- |
| 200 | link to patron info in ILS | "card-granted" | 123456789123456 | username | 1234 | true | "The library card will be a temporary library card. The home address is not in New York State but the work address is in New York City. Visit your local NYPL branch within 30 days to upgrade to a standard card." | 1234567 |
| 200 | link to patron info in ILS | "card-granted" | 123456789123456 | username | 1234 | false | "The library card will be a standard library card." | 1234567 |
| 400 | - | "invalid-request" | - | - | - | - | `{ "address": "Library cards are only available for residents of New York State or students and commuters working in New York City." }` | - |

Example: 
Screen Shot 2020-07-01 at 1 36 54 PM

## Motivation and Context

This resolves [DQ-321](https://jira.nypl.org/browse/DQ-321). Adding a `workAddress` request input allows the user to add more information they need in case they are outside NYS but work in NYC.

## How Has This Been Tested?

Locally through Postman and checking the ILS for the correct type of cards.

## Checklist:


EdwinGuzman:
Here's a document to show all the possible cases for a denied, temporary, or standard card based on address, work address, and policy type - https://docs.google.com/document/d/1wjKvBW8xvT7PDxwp1IMRC2VeGQomsi0OpDikJq8nWA4/edit?usp=sharing

PR #73, opened 2020-07-02
Updating the swagger docs for the new endpoints.

EdwinGuzman:
## Description

The endpoints include `/v0.3/patrons`, `/v0.3/validations/username`, and `/v0.3/validations/address`.

## Motivation and Context

Resolves [DQ-322](https://jira.nypl.org/browse/DQ-322). Makes it easier to review the endpoints for the necessary inputs and returned responses.

## How Has This Been Tested?

On http://editor.swagger.io/ . You can drop in the json or yaml to see the docs BUT you can't actually call the endpoints since it's not authenticated nor on any server.

Screen Shot 2020-07-01 at 9 27 16 PM
Screen Shot 2020-07-01 at 9 27 23 PM

## Checklist:

### DIFF
--- a/api/swagger/swaggerDoc.yaml
+++ b/api/swagger/swaggerDoc.yaml
@@ -73,6 +73,37 @@ paths:
       security:
         - api_auth:
             - 'openid write:patron offline_access api'
+  /v0.3/patrons:
+    x-swagger-router-controller: patrons
+    post:
+      tags:
+        - patrons
+      summary: Create a Patron
+      description: Create a new patron account via the ILS.
+      operationId: patrons_creatorV03
+      parameters:
+        - name: patron
+          in: body
+          description: The information of the new patron
+          required: true
+          schema:
+            $ref: '#/definitions/PatronsCreatorDataV03'
+      responses:
+        '200':
+          description: Successful operation
+          schema:
+            $ref: '#/definitions/PatronsCreatorResponseV03'
+        '400':
+          description: Bad request
+          schema:
+            $ref: '#/definitions/400ErrorResponseV03'
+        '502':
+          description: Generic server error
+          schema:
+            $ref: '#/definitions/502ErrorResponseCreatePatronV03'
+      security:
+        - api_auth:
+            - 'openid write:patron offline_access api'
   /v0.3/patrons/dependents:
     x-swagger-router-controller: patrons
     post:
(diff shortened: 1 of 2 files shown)


api/swagger/swaggerDoc.yaml
```diff
@@ -290,6 +383,45 @@ definitions:
...
+    properties:
+      status:
+        type: number
+        example: 400
+      type:
+        type: string
+        example: invalid-username
+      message:
+        type: object
+        example: 'Usernames should be 5-25 characters, letters or numbers only. Please revise your username.'
```
adriana-alter, line 396: How come this is in quotes?  (The other message strings aren't.)

api/swagger/swaggerDoc.yaml
```diff
@@ -543,6 +726,35 @@ definitions:
...
+  PatronsCreatorDataV03:
+    required:
+      - names
+      - username
+      - pin
+      - address
+    properties:
+      name:
+        type: string
+        example: 'TestFirstName, TestLastName'
```
adriana-alter, line 738: Inconsistent quotes here and throughout the next few lines too.

api/swagger/swaggerDoc.yaml
```diff
@@ -802,6 +1044,22 @@ definitions:
...
         type: object
         example: {}
+  UsernameResponseV03:
+    type: object
+    properties:
+      data:
+        $ref: '#/definitions/UsernameResponseModelV03'
+  UsernameResponseModelV03:
+    properties:
+      type:
```
adriana-alter, line 1054: Is there anything we can call this other than "type"?  It seems confusing given that the data type is also "type"...and then there's something else called "cardType"...

api/swagger/swaggerDoc.yaml
```diff
@@ -813,6 +1071,40 @@ definitions:
         type: boolean
         default: false
         example: true
+  AddressDataV03:
+    description: The data format of the input address
+    required:
+      - address
```
adriana-alter, line 1077: Why the `-`?

api/swagger/swaggerDoc.yaml
```diff
@@ -813,6 +1071,40 @@ definitions:
...
+        type: string
+        default: simplye
+        example: simplye
+  AddressResponseV03:
+    type: object
+    properties:
+      status:
+        type: number
+        example: 200
+      type:
```
adriana-alter, line 1095: And again--can this be renamed?

api/swagger/swaggerDoc.yaml
EdwinGuzman: That's interesting. I don't know. I work with the json file (since I prefer json) and make the Swagger editor convert it to yaml.

api/swagger/swaggerDoc.yaml
EdwinGuzman: Yeah, not sure but it's valid yaml.

api/swagger/swaggerDoc.yaml
EdwinGuzman: It's the standard for good and error messages though. But I see the confusion with `cardType`. The other `type` is part of Swagger but yes, it's a bit confusing.

api/swagger/swaggerDoc.yaml
EdwinGuzman: It's syntax for arrays in yaml.

api/swagger/swaggerDoc.yaml
EdwinGuzman: It's confusing when reading the yaml, but for the API response I think it makes sense to have it there.

EdwinGuzman:
The yaml file is auto-generated after I work with the json file. I kept the yaml file in this repo because some projects have both files, too. But really, it's the json file that gets rendered by the `/docs` API endpoint in AWS to create the overall Swagger doc on https://platformdocs.nypl.org/. So, maybe we can just do away with the yaml since we probably don't need it.

adriana-alter:
Doesn't seem worth investigating whether we can do away with it right now; I'll go ahead and approve.  (Just wanted to make sure there weren't syntax errors that would cause something to explode down the line...but if it's auto-generated, that seems like it will probably be fine...right?)

EdwinGuzman:
yeah, it'll probably be fine... you can also verify by copying/pasting it on http://editor.swagger.io/ and running the interface (actual calls won't go through).

PR #74, opened 2020-07-07
Already validated flag to reduce API calls

EdwinGuzman:
## Description

This makes it so that API calls to the ILS for username availability and to Service Objects for address validation do not get requested if those values have already been validated through the `/api/validations/[username or address]` endpoints.

## Motivation and Context

Resolves [DQ-291](https://jira.nypl.org/browse/DQ-291).

Both the username and the address/work address get validated in the `/patrons` endpoint to create a new account in the ILS. However, it's possible that the username and address/work address have been validated through the `/validations/username` or `/validations/address` endpoints. If that's the case, then instead of making another API call to the ILS or to Service Objects, we can just continue and make one request to the ILS to create the card.

A use case for this is if a client decides to make real-time username or address validations in the form. A patron can get an update on the username or address before submitting the entire form and make necessary updates, if any.

The only new request parameter is `usernameHasBeenValidated` for the `/patrons` endpoint. The address request object already had a `hasBeenValidated` flag. Then it was just a matter of checking the flags and making or not making an API request to the ILS or SO.

## How Has This Been Tested?

Locally through Postman.

## Checklist:

### DIFF
--- a/api/controllers/v0.3/endpoints.js
+++ b/api/controllers/v0.3/endpoints.js
@@ -323,6 +323,7 @@ async function checkAddress(req, res) {
  * @param {HTTP response} res
  */
 async function createPatron(req, res) {
+  const usernameHasBeenValidated = req.body.usernameHasBeenValidated || false;
   let address = req.body.address
     ? new Address(req.body.address, soLicenseKey)
     : undefined;
@@ -336,6 +337,7 @@ async function createPatron(req, res) {
     address: address, // created above
     workAddress: workAddress,
     username: req.body.username, // from req
+    usernameHasBeenValidated,
     pin: req.body.pin, // from req
     email: req.body.email, // from req
     birthdate: req.body.birthdate, // from req
--- a/api/controllers/v0.3/endpoints.js
+++ b/api/controllers/v0.3/endpoints.js
@@ -323,7 +323,6 @@ async function checkAddress(req, res) {
  * @param {HTTP response} res
  */
 async function createPatron(req, res) {
-  const usernameHasBeenValidated = req.body.usernameHasBeenValidated || false;
   let address = req.body.address
     ? new Address(req.body.address, soLicenseKey)
     : undefined;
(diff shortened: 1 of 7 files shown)


api/controllers/v0.3/endpoints.js
```diff
@@ -323,6 +323,7 @@ async function checkAddress(req, res) {
  * @param {HTTP response} res
  */
 async function createPatron(req, res) {
+  const usernameHasBeenValidated = req.body.usernameHasBeenValidated || false;
```
adriana-alter, line 326: Isn't `usernameHasBeenValidated` just a boolean value?  If so, why do you need the `|| false`?  `const usernameHasBeenValidated = !!req.body.usernameHasBeenValidated` seems more readable.

api/controllers/v0.3/endpoints.js
```diff
@@ -336,6 +337,7 @@ async function createPatron(req, res) {
     address: address, // created above
     workAddress: workAddress,
     username: req.body.username, // from req
+    usernameHasBeenValidated,
```
adriana-alter, line 340: Why doesn't this need a key as well as a value?  Also, I'm not sure you need a separate variable for `usernameHasBeenValidated`; it's a little confusing to read, given that almost everything else in here is just `req.body`.whatever.

api/models/v0.3/modelCard.js
```diff
@@ -166,6 +175,7 @@ class Card {
     this.address = this.getOrCreateAddress(args["address"]);
     this.workAddress = this.getOrCreateAddress(args["workAddress"]);
     this.username = args["username"];
+    this.usernameHasBeenValidated = args["usernameHasBeenValidated"] || false;
```
adriana-alter, line 178: Again, if it's just a boolean, `!!args["usernameHasBeenValidated"]` would be more readable...

api/controllers/v0.3/endpoints.js
EdwinGuzman: Yeah, that's a good update.

api/controllers/v0.3/endpoints.js
EdwinGuzman: Yeah, I can clean that up.

PR #78, opened 2020-07-09
Updated readme

EdwinGuzman:
## Description

This updates the README to make it clearer that the `v0.3` endpoints are the stable ones to use. I moved the older endpoint documentation to a separate file. Also made more links to the more extensive documentation in the wiki.

## Motivation and Context

Resolves [DQ-331](https://jira.nypl.org/browse/DQ-331).

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,18 @@
 ## Change Log
 
+### v0.6.0
+
+#### Add
+
+- Added endpoints `validations/address` and `/patrons` for the v0.3 version.
+
+#### Updated
+
+- Updated the address validations to include a work address as well as implementing validation through Service Objects.
+- Updated some of the business logic surrounding the three different policies that can used.
+- Updated the response objects returned for the v0.3 API endpoints to be more consistent.
+- Updated the README to reflect that the v0.3 endpoints are stable and linked to further documentation in the wiki for this project.
+
 ### v0.5.0
 
 #### Add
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,18 @@
...
+### v0.6.0
+
+#### Add
+
+- Added endpoints `validations/address` and `/patrons` for the v0.3 version.
+
+#### Updated
+
+- Updated the address validations to include a work address as well as implementing validation through Service Objects.
+- Updated some of the business logic surrounding the three different policies that can used.
```
adriana-alter, line 12: Missing word--should be "can BE used"

LEGACY_ENDPOINTS_README.md
```diff
@@ -0,0 +1,179 @@
...
+The request data format should be in JSON with required fields of "name", "dateOfBirth", "address", "username", and "pin". The username must be unique. Example data for API v0.1:
+
+```javascript
+{
+  "simplePatron": {
+    "name": "Mikey Olson, Jr.",
+    "dateOfBirth": "11/11/1987",
+    "email": "mjolson@example.com",
+    "address": {
+      "line_1": "123 Fake Street",
```
adriana-alter, line 53: Can this be the same fake address as you have for the other examples below?  (No big deal, obviously--I think it's just a little clearer if the same data is consistently used for all the examples.)

EdwinGuzman: Yes, of course. But this address is used throughout this file only. The other readme file has it's own address it uses. I think it's fine since they are two different files and two different endpoints. Unless if you meant something else?

adriana-alter: No, I meant consistency across both files would be nice...but again, this is not a big deal and I doubt anyone else will ever notice it!

EdwinGuzman: Ah, sure I'll do that.

PR #88, opened 2020-08-18
Adding encrypted database values and decrypted them when the app loads.

EdwinGuzman:
## Description

This adds encrypted database values for the NYPL AWS QA and production environments. They are encrypted through AWS KMS and must be decrypted by KMS.

## Motivation and Context

Resolves [DQ-362](https://jira.nypl.org/browse/DQ-362).

Every time I deploy to QA or production, I manually copy over the database credentials which is annoying. Adding them in the app will automatically update the value in the AWS console and the app will handle decrypting the credentials.

## How Has This Been Tested?

Locally.

## Checklist:


app.js
```diff
@@ -1,32 +1,55 @@
...
-// table and inserting the seed data if it's not in the database already.
-db.init();
+const pathName = `${process.cwd()}/config/deploy_${app.get("env")}.env`;
+require("dotenv").config({ path: pathName });
+
+const BarcodeDb = require("./db");
+
+// Decrypt any database credentials if on QA or production for the NYPL AWS
+// environment. Then create the single database instance in the app.
+async function initDatabase() {
```
EdwinGuzman, line 21: This is the new part, everything else is mostly eslint/prettier doing its thing.

app.js
```diff
@@ -1,32 +1,55 @@
...
+    // These are constant and don't need to be encrypted or decrypted.
+    database: process.env.DB_DATABASE,
+    port: process.env.DB_PORT,
+  });
+  // Once we get the database class instance, initialize it by creating the
+  // table and inserting the seed data if it's not in the database already.
+  db.init();
+}
+
+initDatabase();
```
EdwinGuzman, line 52: I had to make this a function because it's easier to get the asynchronous decrypted values through an async/await in a function.

EdwinGuzman:
Yeah, it's a pain. Unfortunately, I might not be able to do the same for the VPC configuration which is even more of a pain.

PR #95, opened 2020-09-15
Juvenile last name

EdwinGuzman:
## Description

Juvenile cards are created without last names! This fixes that.

## Motivation and Context

Fixes [DQ-369](https://jira.nypl.org/browse/DQ-369). The SimplyE mobile clients mark the child's last name in the juvenile account form as optional... so when the user doesn't enter anything, the account has no last name. The assumption is that the child's last name should be the last name of the parent account.

## How Has This Been Tested?

Through Postman:
Created an account with the parent name "Fortitude Patience" (the ILS stores names as "LASTNAME, FIRSTNAME"). When creating a child account with name "Tommy", the full name is now "Tommy PATIENCE". All names get set to uppercase in the ILS so there is a discrepancy to the user since they see the data as it is submitted but not as it is stored.

Screen Shot 2020-09-15 at 6 35 15 PM

## Checklist:

### DIFF
--- a/api/controllers/v0.3/endpoints.js
+++ b/api/controllers/v0.3/endpoints.js
@@ -17,7 +17,7 @@ const {
   errorResponseDataWithTag,
 } = require("../../helpers/responses");
 const DependentAccountAPI = require("./DependentAccountAPI");
-const { strToBool } = require("../../helpers/utils");
+const { updateJuvenileName } = require("../../helpers/utils");
 
 const ROUTE_TAG = "CREATE_PATRON_0.3";
 // This returns a function that generates the error response object.
@@ -560,10 +560,12 @@ async function createDependent(req, res) {
     ...formattedAddress,
     hasBeenValidated: true,
   });
+  const childsName = updateJuvenileName(req.body.name, parentPatron.names);
+
   // This new patron has a new ptype.
   const policy = Policy({ policyType: "simplyeJuvenile" });
   const card = new Card({
-    name: req.body.name, // from req
+    name: childsName, // from req
     username: req.body.username, // from req
     pin: req.body.pin, // from req
     birthdate: req.body.birthdate, // from req
--- a/api/helpers/utils.js
+++ b/api/helpers/utils.js
@@ -29,7 +29,7 @@ const strToBool = (str) => {
 
 /**
  * updateJuvenileName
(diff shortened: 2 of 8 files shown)


api/controllers/v0.3/endpoints.js
```diff
@@ -17,7 +17,7 @@ const {
   errorResponseDataWithTag,
 } = require("../../helpers/responses");
 const DependentAccountAPI = require("./DependentAccountAPI");
-const { strToBool } = require("../../helpers/utils");
```
adriana-alter, line 20: What was this doing; why don't we need it anymore?

api/helpers/utils.js
```diff
@@ -28,6 +27,38 @@ const strToBool = (str) => {
   return valsHash[found || str.toLowerCase()];
 };
 
+/**
+ * updateJuvenileName
+ * Update the juvenile's name in case no last named was passed with the
```
adriana-alter, line 32: Typo; "last name", not "named"

api/helpers/utils.js
```diff
@@ -28,6 +27,38 @@ const strToBool = (str) => {
...
+const updateJuvenileName = (name, parentArrayName = []) => {
+  const parentsName = parentArrayName[0];
+  // If there's no parent name, then just return the name for the child.
+  if (!parentsName) {
+    return name;
+  }
+
+  let updatedName = name;
+  // If there's no last name, then use the parent's last name. This is a very
+  // basic check that is done by checking if there is a space in the complete
```
adriana-alter, line 46: But it's possible for someone's first name to be more than one word...

api/controllers/v0.3/endpoints.js
EdwinGuzman: If the client passes booleans as strings, `"false"` or `"true"`, this converts it to the boolean value, but it wasn't used in this file. Just accidentally imported.

api/helpers/utils.js
EdwinGuzman: Yes, but right now we don't take first or last names as separate request inputs. It's just `"name"`. For now, this is fine but I'll make a ticket to make the request accept first and last name which is a larger update than this which is needed soon.

adriana-alter (approved):

PR #99, opened 2020-10-09
Updating the error responses

EdwinGuzman:
## Description

Trying to update error responses to be more like a problem detail. The endpoints should now return `status`, `type`, `title`, and `detail` along with any specific properties like `cardType`. The `/dependents` and `/dependents-eligibility` endpoints are used by the SimplyE clients so they still retain the `name` and `message` properties but also include `title` and `detail`. So there are some duplicates. They will be removed once SimplyE updates or stops using the endpoints (much later on down the road).

I was also able to remove and clean some of the code that converts errors into json. Javascript `Error`s mainly uses the `.message` property when displaying in the console or terminal so you'll see that those classes still have `.message` in there. It's also needed for testing to make sure that a function throws the right error message.

But, once it gets to the client, those js `Error`s are converted into json objects and the `.message` is turned into `.detail` (with the exception of some errors for the dependent endpoints clients use and still need `.message`).

I must admit I lost my head around this conversion so I will further clean it up. I also added a jest config file to not test any of the `0.1` and `0.2` tests since those endpoints are deprecated.

## Motivation and Context

Resolves [DQ-414](https://jira.nypl.org/browse/DQ-414). When attempting to update the error handling in the front-end, I noticed this was more complicated in the back-end. So this is an attempt to get a more standard approach. There are still some hiccups here and there with redundant messages in two properties just to appease any existing clients depending on those properties. 

## How Has This Been Tested?

Locally and through Postman. Here are some examples:

Screen Shot 2020-10-08 at 12 21 20 PM

Screen Shot 2020-10-08 at 3 24 22 PM

Example of duplicate messages. This will be further cleaned up later but is mostly used to cover and backwards-compatibility for SimplyE.

Screen Shot 2020-10-08 at 5 19 43 PM

Specific errors also display better:
Screen Shot 2020-10-08 at 3 58 56 PM

## Checklist:

### DIFF
--- a/api/controllers/v0.3/IlsClient.js
+++ b/api/controllers/v0.3/IlsClient.js
@@ -54,14 +54,14 @@ const IlsClient = (args) => {
       })
       .catch((error) => {
         const response = error.response;
+        const message =
+          response.data && (response.data.description || response.data.name);
 
         // If the request to the ILS is missing a value or a key is of
         // and incorrect type, i.e. the barcode is sent as an integer
         // instead of a string.
         if (response.status === 400) {
-          throw new InvalidRequest(
-            `Invalid request to ILS: ${response.data.description}`
-          );
+          throw new InvalidRequest(`Invalid request to ILS: ${message}`);
         }
 
         if (!(response.status >= 500)) {
@@ -158,8 +158,10 @@ const IlsClient = (args) => {
         return response;
       })
       .catch((error) => {
-        logger.error(`Error calling ILS - ${error.response}`);
+        const data = error.response && error.response.data;
+        const message = data.description || data.name || "Unknown Error";
         logger.error(`Error calling ILS URL - ${findUrl}${params}`);
+        logger.error(`Error calling ILS - ${message}`);
         return error.response;
       });
 
--- a/api/controllers/v0.3/UsernameValidationAPI.js
+++ b/api/controllers/v0.3/UsernameValidationAPI.js
@@ -1,11 +1,10 @@
-/* eslint-disable */
 const { NoILSClient, BadUsername } = require("../../helpers/errors");
 
 /**
  * A class that validates usernames against the ILS.
  */
 const UsernameValidationAPI = (args) => {
-  const ilsClient = args["ilsClient"];
+  const ilsClient = args.ilsClient;
   const USERNAME_PATTERN = /^[a-zA-Z0-9]{5,25}$/;
   const AVAILABLE_USERNAME_TYPE = "available-username";
   const UNAVAILABLE_USERNAME_TYPE = "unavailable-username";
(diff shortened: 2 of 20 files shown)


api/controllers/v0.3/IlsClient.js
```diff
@@ -54,14 +54,14 @@ const IlsClient = (args) => {
       })
       .catch((error) => {
         const response = error.response;
+        const message =
+          response.data && (response.data.description || response.data.name);
```
EdwinGuzman, line 58: Depending on the error, the ILS returns different error objects.

api/controllers/v0.3/UsernameValidationAPI.js
```diff
@@ -30,30 +29,12 @@ const UsernameValidationAPI = (args) => {
...
   };
 
-  /**
-   * validate(username)
-   * Checks if the username is valid and available and returns and object
-   * with the appropriate response.
-   *
-   * @param {string} username
-   */
-  const validate = async (username) => {
```
EdwinGuzman, line 40: This function was just moved below.

api/controllers/v0.3/endpoints.js
```diff
@@ -127,19 +128,11 @@ async function setupEndpoint(endpointFn, req, res) {
     soLicenseKey =
       soLicenseKey || (await awsDecrypt.decryptKMS(process.env.SO_LICENSE_KEY));
   } catch (error) {
-    const localErrorMessage =
-      "The ILS client key and/or secret, or the Service Objects license key were not decrypted";
-
-    const errorResponseData = modelResponse.errorResponseData(
```
EdwinGuzman, line 133: This was mostly redundant and now I think using just one function, `collectErrorResponseData` with the error object, is easier to understand. It just convert the error object and also logs it.

api/controllers/v0.3/endpoints.js
```diff
@@ -397,15 +393,10 @@ async function createPatron(req, res) {
...
-      ...modelResponse.errorResponseData(
-        collectErrorResponseData(
-          400,
-          "invalid-request",
-          "There was an error with the request.",
-          "",
-          errors
-        ) // eslint-disable-line comma-dangle
+      ...collectErrorResponseData(
+        new InvalidRequest("There was an error with the request.")
```
EdwinGuzman, line 397: There are still some places where the error is not coming from a controller function but rather directly here. Eventually I want to update these.

api/helpers/errors.js
```diff
@@ -1,208 +1,297 @@
...
+class InvalidRequest extends ProblemDetail {
+  constructor(detail) {
+    super();
+    this.status = 400;
     this.type = "invalid-request";
-    this.state = 400;
-    this.message = message;
+    this.title = "Invalid Request";
+    this.message = detail;
+    // To support older versions of API where client expect these values:
```
EdwinGuzman, line 51: All error objects will render `title` and `message` so it can be a standard javascript `Error` object. The `message` gets converted into a `detail` when it's returned to the client. In some cases, we _still_ want to return `name` and `message` to the client and that's what happens in the following lines. 

This gets repeated in other error classes.

api/helpers/validateEnvironment.js
```diff
@@ -1,18 +1,15 @@
-const isEmpty = require('underscore').isEmpty;
```
EdwinGuzman, line 1: I never got a chance to clean this up. So now this is cleaner and a lot of unneeded or unnecessary code is removed.

tests/unit/models/v0.3/modelCard.test.js
```diff
@@ -155,7 +157,7 @@ describe("CardValidator", () => {
       card.address.validate = oldValidate;
     });
 
-    it("should update the errors object in the card if any errors are returned", async () => {
+    it.skip("should update the errors object in the card if any errors are returned", async () => {
```
EdwinGuzman, line 160: This is a separate issue but not necessarily a bug. If Service Objects throws an error or says an address is invalid for the `/validations/address` endpoint, we return that message. For the `/patrons` endpoint, however, if there's an error validating an address, we want to continue anyway and do a naive check on the city and state fields. This will return a temporary card to the user but it won't return an error. This is business logic planned out since Service Objects can be down for whatever reason.

https://jira.nypl.org/browse/DQ-415

adriana-alter (approved):

PR #101, opened 2020-10-19
New barcode values

EdwinGuzman:
## Description

New p-types will be introduced soon. Those p-types will get barcodes that start with a new sequence, so the model needs to be able to generate a new barcode based on the barcode sequence type, *not* the p-type. The `Barcode` and `BarcodesDb` classes shouldn't know or care about p-types. (That connection between p-types and barcode starting sequence will be implemented in a later PR).

## Motivation and Context

Resolves [DQ-418](https://jira.nypl.org/browse/DQ-418). New p-types will need a new barcode scheme.

## How Has This Been Tested?

Locally.

## Checklist:

### DIFF
--- a/db/index.js
+++ b/db/index.js
@@ -48,19 +48,26 @@ class BarcodesDb {
 
   /**
    * initInsert()
-   * Insert the first value into the database if it's not already there. This is
-   * to set from where new barcodes will be created.
+   * Insert the first values into the database if they are not already there.
+   * This is to set from where new barcodes will be created based on p-types.
    */
   async initInsert() {
     const text = "INSERT INTO barcodes (barcode, used) VALUES ($1, $2);";
-    const values = ["28888855432452", "true"];
+    const barcodes = ["28888855432452", "25555001345283"];
 
-    try {
-      await this.pool.query(text, values);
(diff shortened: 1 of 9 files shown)


db/index.js
```diff
@@ -48,19 +48,28 @@ class BarcodesDb {
...
    * initInsert()
-   * Insert the first value into the database if it's not already there. This is
-   * to set from where new barcodes will be created.
+   * Insert the first values into the database if they are not already there.
+   * This is to set from where new barcodes will be created based on p-types.
    */
   async initInsert() {
     const text = "INSERT INTO barcodes (barcode, used) VALUES ($1, $2);";
-    const values = ["28888855432452", "true"];
+    const barcodes = ["28888855432452", "25555001345283"];
```
EdwinGuzman, line 56: This is the main part and introduction of the new barcode. The insertion into the db is done in a `Promise.all` to make sure it successfully or unsuccessfully runs for both barcodes.

adriana-alter (approved):
Nice!!

EdwinGuzman:
Thanks. There's another barcode to add but it won't be until December (I think).

PR #102, opened 2020-10-20
Policy and IlsClient class refactor

EdwinGuzman:
## Description

This refactors the `Policy` and `IlsClient` classes. Nothing new is actually added but I think it's cleaner and it fixed a few things I didn't get a chance to refactor before.
This also adds updates eslint which I know should've been a separate PR, but some syntax issues in my IDE were annoying me.

Updates:
* Before, the main `endpoints` file took care of generating the `ilsToken` to make authenticated calls to the ILS. If the token was expired, it would generate a new one. It did all of this in the global scope of the file for all endpoints. This wasn't an issue but all ILS-related functions and variables are now inside the instance of the `IlsClient` class. The token is now private in the class so the app doesn't know about it in `endpoints.js` but all endpoint functions can still make authenticated calls. This was the biggest update and makes the code better since some endpoints like the address validation endpoint doesn't care about the ILS token. The `IlsClient` instance creation only happens _once_ when the app is created and used throughout the app. The token is also generated once at first and again only when it's expired. This hasn't changed but pointing out that each new API request does not create a new `IlsClient` instance nor does it generate a new token.
* Added more tests especially for this ILS token API call.
* The `Policy` class was updated so that each ptype now has its own standard and temporary expiration time policy. Some ptypes overlap and that's okay, they're all combined in a `switch` statement. This is because upcoming ptypes have different expiration times and they might have different policies. So now the expiration time and policy type are not as coupled, it's the expiration time with the ptype.
* Minor logic was updated in the `Card` class so that ptypes are set whenever a card is validated instead of setting the ptype immediately before calling the ILS to create an account. This was done because now the ptype is coupled with the expiration date and we want to know the date after making sure the data is validated.
* Resulting ILS account creation hasn't changed, just the internals.

## Motivation and Context

Resolves [DQ-420](https://jira.nypl.org/browse/DQ-420). This is to make it easier to add new ptypes with separate expiration dates.

## How Has This Been Tested?

Locally and through tests.

## Checklist:


package.json
```diff
@@ -41,25 +41,26 @@
...
   "devDependencies": {
     "@babel/preset-env": "^7.9.5",
+    "babel-eslint": "^10.1.0",
     "babel-jest": "^25.4.0",
     "chai": "4.2.0",
     "chai-as-promised": "^7.1.1",
-    "eslint": "7.1.0",
-    "eslint-config-airbnb": "^18.1.0",
-    "eslint-config-airbnb-base": "14.1.0",
+    "eslint": "7.11.0",
```
EdwinGuzman, line 56: I accidentally committed these new updates to this branch but it's just minor package updates to fix security issues.

adriana-alter (approved):

PR #104, opened 2020-11-02
P-Type refactor

EdwinGuzman:
## Description

This updates the logic behind the newly added web digital ptypes (7, 8, 9). But, in order to get the logic right, I had to refactor a lot of the code so this PR is very big. Sorry :( .

This does clean up a lot of things that were implemented in a way that could've been better and now, with these ptypes and updated policies, it was easier to make those changes.

The logic for the code goes like this:
* `Address` class -> this now has methods to check if the current address is in NYC, NYS, or in the US.
* `Card` class -> this uses the methods from the Address class for both the home and work address (the `Card` class cares whether the address is a home or work address and the `Address` class does not). The `CardValidator` class is now part of the `Card` class. I think it makes sense to combine and not sure why it was separate before. The validation process is (hopefully) easier to follow. The basic values are checked and then the username and address are validated. Returned responses are also updated. Now, when we run `card.validate`, if the card values are all valid, it will attempt to set the ptype instead of having to call that method separately. There is no longer the idea of a "temporary" and "standard" card. All cards are the same but what differentiates them is the expiration time. For the `setBarcode` method, we now have to give different ptypes different barcodes... so now this checks for the ptype of the account before attempting to generate a new barcode. The new web digital ptypes (7, 8, 9) get barcodes that start with "25555" and the simplyeJuvenile ptype (4) get barcodes that start with "288888" (this was done in a previous PR).
* `Policy` class -> this has an updated set expiration times and policy types and ptypes. This also calls the location based methods from the `Card` class to determine the ptype. **Note**: this is actually the part of the JIRA ticket. The logic for the new ptypes are in `determinePtype`. This is the same logic described in the JIRA ticket so I won't write it down here.

`webApplicant` is now the default policy type and we will implement `simplye` later on. `simplyeJuvenile` still works as before.
The rest of the update is fixing and cleaning up tests.

## Motivation and Context

Resolves [DQ-374](https://jira.nypl.org/browse/DQ-374). There may be an update to this later but this is okay for now so it doesn't get any bigger.

## How Has This Been Tested?

Locally and created the different ptypes in the QA ILS server.

## Checklist:


api/helpers/utils.js
```diff
@@ -89,8 +89,87 @@ const updateJuvenileName = (name, parentArrayName = []) => {
   return updatedName;
 };
 
+const nyCities = ["New York", "New York City", "NYC"];
+const nyCounties = ["Richmond", "Queens", "New York", "Kings", "Bronx"];
+const nyStates = ["NY", "New York"];
+const listOfStates = [
```
adriana-alter, line 95: This is fine, but there are packages that would probably do a lot of this for you (maybe `https://www.npmjs.com/package/usa-states`), without having to manually list everything.

api/helpers/utils.js
```diff
@@ -89,8 +89,87 @@ const updateJuvenileName = (name, parentArrayName = []) => {
   return updatedName;
 };
 
+const nyCities = ["New York", "New York City", "NYC"];
+const nyCounties = ["Richmond", "Queens", "New York", "Kings", "Bronx"];
+const nyStates = ["NY", "New York"];
```
adriana-alter, line 94: Why do we need the options--are users allowed to enter the states in multiple formats, or...?

api/models/v0.3/modelPolicy.js
```diff
@@ -1,67 +1,40 @@
 const IlsClient = require("../../controllers/v0.3/IlsClient");
 
-const lowerCase = (arr) => arr.map((item) => item.toLowerCase());
-
 /**
  * Creates a policy object to find out what type of card is allowed for a
- * given patron and their location.
+ * given card and their location.
```
adriana-alter, line 5: Why the change from "patron" to "card" here?

api/helpers/utils.js
EdwinGuzman: I didn't want to bring in a package for this and this seemed the simplest solution.

api/helpers/utils.js
EdwinGuzman: True, I'll remove "New York" because that did give me issues.

api/models/v0.3/modelPolicy.js
EdwinGuzman: Because the class is called `Card`. I didn't make updates elsewhere but I can fix that naming issue in a later PR.

EdwinGuzman: I did make that mistake from the Ruby Card Creator where it used Patron and Card interchangeably and it would be better to just use `Card`.

adriana-alter (approved):

PR #107, opened 2020-11-13
Minor refactor and cleanup for unit tests.

EdwinGuzman:
## Description

Also, cleaned up comments and parameters for some classes and functions.

## Motivation and Context

Resolves [DQ-287](https://jira.nypl.org/browse/DQ-287). Just wanted to clean up tests and update tests that used the `simplye` policy before when now the `webApplicant` policy is the default/main type.

## How Has This Been Tested?

Locally.

## Checklist:

### DIFF
--- a/api/controllers/v0.3/AddressValidationAPI.js
+++ b/api/controllers/v0.3/AddressValidationAPI.js
@@ -1,18 +1,16 @@
-/* eslint-disable */
 const ServiceObjectsClient = require("./ServiceObjectsClient");
 const { SONoLicenseKeyError } = require("../../helpers/errors");
 const { strToBool } = require("../../helpers/utils");
 
 /**
  * A class that uses Service Objects to validate addresses.
+ * @param {string} soLicenseKey License key for the Service Objects API.
  */
-const AddressValidationAPI = (args = {}) => {
-  const soLicenseKey = args["soLicenseKey"];
-
+const AddressValidationAPI = (soLicenseKey) => {
   const ALTERNATE_ADDRESSES_TYPE = "alternate-addresses";
   const UNRECOGNIZED_ADDRESS_TYPE = "unrecognized-address";
   const VALID_ADDRESS_TYPE = "valid-address";
-  const { validateAddress } = ServiceObjectsClient({ soLicenseKey });
+  const { validateAddress } = ServiceObjectsClient(soLicenseKey);
 
   const RESPONSES = {
     unrecognized_address: {
@@ -30,7 +28,7 @@ const AddressValidationAPI = (args = {}) => {
   };
 
   /**
-   * validate(address)
+   * validate
    * Calls the Service Objects client to validate an address and returns a
    * response with the type of address it is, even if SO is not callable.
    * @param {Object} address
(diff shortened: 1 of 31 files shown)


api/controllers/v0.3/AddressValidationAPI.js
```diff
@@ -1,18 +1,16 @@
...
 const { strToBool } = require("../../helpers/utils");
 
 /**
  * A class that uses Service Objects to validate addresses.
+ * @param {string} soLicenseKey License key for the Service Objects API.
  */
-const AddressValidationAPI = (args = {}) => {
-  const soLicenseKey = args["soLicenseKey"];
-
+const AddressValidationAPI = (soLicenseKey) => {
```
EdwinGuzman, line 9: For some classes, the object prop only had on attribute so it's just easier to pass that argument itself rather than in an object.

api/controllers/v0.3/AddressValidationAPI.js
```diff
@@ -60,11 +58,11 @@ const AddressValidationAPI = (args = {}) => {
   };
 
   /**
-   * createAddressFromResponse(address)
+   * createAddressFromResponse
```
EdwinGuzman, line 61: And these updates are for consistency.

package.json
```diff
@@ -51,8 +51,6 @@
     "@babel/preset-env": "^7.9.5",
     "babel-eslint": "^10.1.0",
     "babel-jest": "^25.4.0",
-    "chai": "4.2.0",
-    "chai-as-promised": "^7.1.1",
```
EdwinGuzman, line 55: These aren't needed.

tests/unit/controllers/v0.3/AddressValidationAPI.test.js
```diff
@@ -337,27 +321,6 @@ describe("AddressValidationAPI", () => {
       });
     });
 
-    it("returns a 'valid-address' response but card denied if address is outside NY", () => {
```
EdwinGuzman, line 340: We are no longer denying accounts if a patron is outside NY. They just get a temporary account.

tests/unit/controllers/v0.3/DependentAccountAPI.test.js
```diff
@@ -234,12 +253,26 @@ describe("DependentAccountAPI", () => {
     });
   });
 
+  describe("checkAccountExpiration", () => {
```
EdwinGuzman, line 256: Some tests aren't new but I just moved them around so it appears as if they are new. This is just to match up the order of the function declaration in the class' file. It just makes it easier to read for me when the function and tests for that function are in the same order.

tests/unit/models/v0.3/modelCard.test.js
```diff
@@ -1081,64 +1050,63 @@ describe("Card", () => {
...
-      ...basicCard,
-      policy: Policy({ policyType: "simplye" }),
-    });
-    const webCard = new Card({
+    const card = new Card({
       ...basicCard,
       policy: Policy({ policyType: "webApplicant" }),
     });
 
-    it("should set the temporary expiration date to 30 days for simplye policy", () => {
```
EdwinGuzman, line 1093: There's no longer a "standard" and a "temporary" expiration date for a ptype. Each ptype only has one expiration date now.

api/controllers/v0.3/AddressValidationAPI.js
```diff
@@ -30,7 +28,7 @@ const AddressValidationAPI = (args = {}) => {
   };
 
   /**
-   * validate(address)
+   * validate
```
adriana-alter, line 31: Why not keep the arguments in the comments?  Makes it easier to understand what the function does...

EdwinGuzman: I felt it was redundant since the comment has the `@param` example and then the function definition is right below it. I'm not sure what the "right" format is; I've seen both examples in multiple repos but without the param seems more common.

adriana-alter (approved):

PR #110, opened 2020-12-07
Making sure there's an ILS token before proceeding on any v0.3 endpoint.

EdwinGuzman:
## Description

There was an issue with the ILS token generation when a new Lambda started up. It is an async function but I forgot to add `await`. This caused a race condition where the ILS was called but there was no token! That caused a few unauthenticated ILS API calls and caused errors when creating new users.

That was fixed but this error _still_ comes up in production, but just not as often as before. I can't say I know exactly why this happens but it's mostly when a new lambda instance spins up. In order to fix this, I'm just making sure that the ILS token check - if it is there and not expired - succeeds by calling it twice. There's no problem just checking if everything is okay so there's no loss in speed when calling the endpoints.

## Motivation and Context

Resolves [DQ-447](https://jira.nypl.org/browse/DQ-447).

## How Has This Been Tested?

Locally with prod env setup.

## Checklist:

### DIFF
--- a/api/controllers/v0.3/endpoints.js
+++ b/api/controllers/v0.3/endpoints.js
@@ -47,7 +47,9 @@ const envVariableNames = [
 /**
  * checkIlsToken
  * check to see if the ILS token is available or if it is expired. If either is
- * true, then request a new one. If there was an issue, return an error.
+ * true, then request a new one. If there was an issue, return an error. Returns
+ * an array [boolean, message] with information if the ILS token generation
+ * call was successful or not.
  */
 async function checkIlsToken(req, res) {
   // If the ilsClient has no token or the token is expired, then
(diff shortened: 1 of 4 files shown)


api/controllers/v0.3/endpoints.js
```diff
@@ -44,6 +44,30 @@ const envVariableNames = [
...
+ * an array [boolean, message] with information if the ILS token generation
+ * call was successful or not.
+ */
+async function checkIlsToken(req, res) {
+  // If the ilsClient has no token or the token is expired, then
+  // generate a new token.
+  if (!ilsClient.hasIlsToken() || ilsClient.isTokenExpired()) {
+    try {
+      await ilsClient.generateIlsToken();
+      return [false, null];
```
adriana-alter, line 60: Maybe it's worth returning a more informative message there?  (Or will no one ever see it?)

EdwinGuzman: Right, no one will ever see it. If the first value is true, then the second value is returned.

adriana-alter (approved):

PR #113, opened 2021-01-06
Lamba VPC information for AWS

EdwinGuzman:
## Description

This adds VPC information from the QA and production AWS environments to the deployment script. Previously, every time Travis deployed the lambda to AWS, the VPC information was lost and I had to add it manually. Not only was this tedious, but it also meant that if it wasn't configured then the lambda couldn't talk to the API Gateway or to the RDS database. More importantly, when new devs work on this and make a deployment, they should not need to worry about configuring the lambda correctly.

To fix this, the VPC subnets and security groups were added to the QA and production `node-lambda` scripts. Now, whenever there's a deployment to QA or production, the VPC information is configured automatically.

## Motivation and Context

Resolves [DQ-451](https://jira.nypl.org/browse/DQ-451). It's annoying to update the VPC information every time there's a new deployment.

## How Has This Been Tested?

Tested by deploying the lambda to the QA environment and making sure that the VPC information was configured correctly.

## Checklist:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ deploy:
   skip_cleanup: true
   script: npm run deploy-qa
   on:
-    branch: qa
+    branch: qa-vpc
 - provider: script
   skip_cleanup: true
   script: npm run deploy-production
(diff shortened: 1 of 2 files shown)


adriana-alter (approved):
Nice!!

PR #114, opened 2021-01-14
Name input fix

EdwinGuzman:
## Description

A staff member from AskNYPL found out that when middle names were entered into the form, the last name was dropped and the middle name was used as the last name. So something like "Bart Jojo Simpson" turned into "JOJO, BART" in the ILS.

This fixes that bug and also handles the case where multiple last names are entered, with the assumption that the first two names in the string are the first and middle names. A name like "Albert Bruce Carlos Don Ellis" will turn into "CARLOS DON ELLIS, ALBERT BRUCE" where "CARLOS DON ELLIS" is the "last name" and "ALBERT BRUCE" is the "first name" (as it's broken down by the ILS.

## Motivation and Context

Fixes [DQ-458](https://jira.nypl.org/browse/DQ-458).

## How Has This Been Tested?

Locally and checked in the QA ILS:
Screen Shot 2021-01-14 at 4 18 54 PM

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
 - Fixed issue where middle names were being used as last names and the last name was completely omitted.
 - Added subnets and security groups for the node-lambda deployment script.
 
+#### Updated
+
+- Updated the problem detail error objects for the Dependent Eligibility endpoint to distinguish between the "limit reached" and "not eligible" errors.
+
 ### v0.7.9
 
 #### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 ## Change Log
 
+### v0.8.0
+
+#### Fixed
+
+- Fixed issue where middle names were being used as last names and the last name was completely omitted.
+- Added subnets and security groups for the node-lambda deployment script.
+
 ### v0.7.9
 
 #### Fixed
--- a/README.md
+++ b/README.md
@@ -159,7 +159,6 @@ Example request:
 {
   "usernameHasBeenValidated": false,
   "username": "tomnook42",
-  "name": "Tom Nook",
   "firstName": "Tom",
   "lastName": "Nook",
   "address": {
--- a/api/controllers/v0.3/DependentAccountAPI.js
+++ b/api/controllers/v0.3/DependentAccountAPI.js
@@ -6,6 +6,7 @@ const {
   NoBarcode,
   ExpiredAccount,
   NotEligibleCard,
+  JuvenileLimitReached,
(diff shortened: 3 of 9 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 ## Change Log
 
+### v0.8.0
+
+#### Fixed
+
+- Fixed issue where middle names were being used as last names and the last name was completely omitted.
+- Added subnets and security groups for the node-lambda deployment script.
```
EdwinGuzman, line 8: I forgot to update this last time but I didn't make a deployment for this, so it's bundled in the same version.

adriana-alter:
The code looks good, but just curious about the rationale for assuming that everything after the first two names is the last name; I realize this is a small and irrelevant sample size, but off the top of my head, I know way more people who have two middle names than people who have a three-word last name...

EdwinGuzman:
I didn't do any research or have numbers to justify that, but it seems to be the most common case. It's definitely cultural too as I know quite a lot of people with first, middle, and two last names.

This hasn't been a problem until a staff member pointed it out when they were testing it, so no patron has complained about this (yet?). The good thing is that regardless of this "fix" or the current implementation, patrons can call in to get their name updated. I'll email back and see how common these errors are and see if they want a specific pattern as well.

EdwinGuzman:
Actually, I'm going to change how the value is passed from the app to this API and from this API to the ILS.

The app sends names in one field as `name`. I'll change it so that the app sends `firstName` and `lastName` and this API will handle the correct format. So if someone enters "Albert Bruce Carlos" in the `first name field` and "Don Ellis Frank" in the `last name field`, the API will format it as "DON ELLIS FRANK, ALBERT BRUCE CARLOS" and send it to the ILS. That way it'll cover multiple middle names and multiple last names (if any).

This is a bit more involved so expect an update next week.

adriana-alter:
That sounds like a great idea!

EdwinGuzman:
Update:

Updated the `normalizeName` function to fix the format of the patron's name. This does the bulk of the work now to get the order correct and prefers the `firstName` and `lastName` arguments so it includes middle name(s) and last name(s). The `formatPatronName` function now just uppercases the full name before it gets sent to the ILS.

Test this end-to-end from the library card app to the QA ILS.

App - used `firstName` = "Edwin Rafael" and `lastName` = "Guzman Carino":
Screen Shot 2021-01-20 at 11 27 04 AM

API data sent to the ILS:
Screen Shot 2021-01-20 at 11 28 12 AM

adriana-alter (approved):
Looks good!!

PR #125, opened 2021-06-16
Updating the logic of the 4-digit pin to an 8-32 character password.

EdwinGuzman:
## Description

This updates the API to take a `password` and tests it against new rules. The changes in the codebase include changing the text pin to password, updating tests, and updating the regular expression that is used to test the input.

Note: the `pin` attribute is still supported for any existing clients that won't change the request data object to include `password` instead of `pin`. Also, even though `password` is used throughout the codebase, the attribute `pin` is _still_ being used when sending the data to the ILS.

## Motivation and Context

Resolves [ERD-14](https://jira.nypl.org/browse/ERD-14).
[Related library card app client PR](https://github.com/NYPL/nypl-library-card-app/pull/128).

Previously, a Library account could have been created with a 4-digit pin. In order to increase security, the ILS team updated this rule to be a password string that is:

* 8-32 alphanumeric characters
* has upper and lower case characters
* has at least one special character

Example: MyLib1731@!

## How Has This Been Tested?

Locally through Postman by making authenticated calls to the QA and production API. A full end-to-end test was done with the Library Card App as discussed in its [PR](https://github.com/NYPL/nypl-library-card-app/pull/128).

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 ## Change Log
 
+### v0.8.3
+
+#### Updated
+
+- Updated the "pin" name attribute to "password" for v0.3 endpoints, but "pin" is still supported for legacy API calls to the endpoints. The ILS API still requires the attribute name "pin". This update changes the format of the existing 4-digit pin to 8-32 alphanumeric password.
+
 ### v0.8.2
 
 #### Updated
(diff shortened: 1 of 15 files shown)


nypl-wluisi (approved):
Code looks good!

PR #126, opened 2021-06-22
Erd 16/password check

EdwinGuzman:
## Description

This removes the regular expression that was checking passwords for uppercase and lowercase letters, at least one number, and at least one special character. The copy in the app was a strong suggestion/encouragement but not actually requirements. The only requirement is for passwords to be between 8 to 32 characters in length.

## Motivation and Context

Resolves [ERD-16](https://jira.nypl.org/browse/ERD-16)

## How Has This Been Tested?

Locally with the front-end [PR](https://github.com/NYPL/nypl-library-card-app/pull/129).

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 ## Change Log
 
+### v0.8.4
+
+#### Fixed
+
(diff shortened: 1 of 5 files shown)


nypl-wluisi (approved):
code looks good!

PR #133, opened 2022-09-21
Encrypted ILS client key and secret for QA

gonuguntla:
## Description

Encrypted ILS client key and secret for QA

## Checklist:


EdwinGuzman (approved):
This works fine.

EdwinGuzman:
BTW, when I say that this works, I only tested the v0.3 endpoints for creating patrons through the ILS v6 API. I did not test the v0.2 endpoints that use the ILS v5 API because I don't know what to look for. I'm assuming you tested this locally for MLN. Is that correct?

PR #157, opened 2025-01-02
rm streaming

charmingduchess:
## Description

Remove streaming of patron creation to NewPatron streams. 

## Motivation and Context

It seems like no one is using it, and there is another kinesis stream with an actively maintained sierra poller which should satisfy the same requirements.

## Checklist:


EdwinGuzman (approved):
The other PR can be pushed to prod soon, but only deploy this to QA for now and leave it there for a week or so to verify no one is using the stream.

PR #164, opened 2025-01-17
prod deploy

charmingduchess:
- rm streaming to NewPatron stream
- add logs for unavailable username 200 responses from sierra

This should wait till Monday for deployment!


EdwinGuzman (approved):
This looks good but wait until Tuesday to deploy to prod.

PR #166, opened 2025-02-21
add log for patron info

charmingduchess:
We are seeing TONS of "Pin is trivial" errors. This log is to verify what people are submitting and if it is really a bad pin. It does contain PII, but the logs are deleted every 4 weeks, and if the patron create function fails, then the info logged does not actually correspond to an actual account being created.

### DIFF
--- a/api/controllers/v0.3/endpoints.js
+++ b/api/controllers/v0.3/endpoints.js
@@ -382,7 +382,8 @@ async function createPatron(req, res) {
           "controller createPatron error calling card.createIlsPatron",
           error
         );
-        logger.info("errored patron info", card);
+        if (error?.message.includes("PIN is trivial"))
+          logger.info("errored patron info", card.password);
         // There was an error hitting the ILS to create the patron. Catch
         // and return the error.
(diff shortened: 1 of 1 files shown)


api/controllers/v0.3/endpoints.js
```diff
@@ -378,14 +378,11 @@ async function createPatron(req, res) {
...
       } catch (error) {
-        console.log(
-          "controller createPatron error calling card.createIlsPatron",
-          error
-        );
         logger.error(
           "controller createPatron error calling card.createIlsPatron",
           error
         );
+        logger.info("errored patron info", card);
```
EdwinGuzman, line 385: Can you be more specific about what's being logged so it doesn't include all the card's information but only the relevant ones?

charmingduchess: I do want to see the whole request that's being sent. I guess we could start with just the pin, but I have a feeling that it could be an issue with the request in general, and not just the pin.

EdwinGuzman (approved):

PR #167, opened 2025-02-27
Move to terraform/gha flow to deploy logs

charmingduchess:
## Checklist:


.github/workflows/test-and-deploy.yml
```diff
@@ -0,0 +1,35 @@
...
+        run: npm ci
+      - name: Run tests
+        run: npm test
+  deploy-qa:
+    runs-on: ubuntu-latest
+    permissions: 
+      id-token: write
+      contents: read
+    needs: test
+    if: github.ref == 'refs/head/qa'
```
EdwinGuzman, line 22: Will the production step come later once this works?

EdwinGuzman (approved):
This looks good. Check out `tests/db/db.test.js` and make sure that line 5, `if (!process.env.TRAVIS)` is not a problem for the tests that run in GHA.

.github/workflows/test-and-deploy.yml
charmingduchess: yea! just dont want this to functionally deploy but be missing config or something like that.

PR #175, opened 2025-08-12
Deploy Sierra creds update

Toxiapo:
## Description

Trigger qa build with new Sierra keys

## Checklist:


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ### v0.9.1
 
 - update web non metro expiration to 3 years [SCC-4658](https://newyorkpubliclibrary.atlassian.net/browse/SCC-4658)
+- Update QA and development Sierra creds
```
EdwinGuzman, line 6: Is the non-metro exp update not deployed yet? If it is, you might need to create a new version or entry in the changelog but it's minor.

PR #192, opened 2026-01-22
SWIS-251: Change default library code from eb to vr

Toxiapo:
## Description

Per comms request, we will be changing our default library `vr` instead of `eb`

https://newyorkpubliclibrary.atlassian.net/browse/SWIS-251

Updates include
1. Ensure the default library code to be `vr`
2. Ensure all incoming `eb` codes to be stored as `vr` 

## Checklist:

### DIFF
--- a/api/helpers/utils.js
+++ b/api/helpers/utils.js
@@ -199,6 +199,20 @@ const normalizedBirthdate = (birthdate) => {
   return;
 };
 
+/**
+ * Temporary update as of Jan 22, 2026.
+ *
+ * We will be changing the default library code for digital cards from "eb" to "vr".
+ * Once the front-end is updated to send "vr" as the homeLibraryCode, we can
+ * remove this mapping function.
+ */
+const mapEbToVr = (code) => {
+  if (code === "eb") {
+    return "vr";
+  }
+  return code;
+};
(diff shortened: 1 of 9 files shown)


api/helpers/utils.js
```diff
@@ -199,6 +199,20 @@ const normalizedBirthdate = (birthdate) => {
...
 
+/**
+ * Temporary update as of Jan 22, 2026.
+ *
+ * We will be changing the default library code for digital cards from "eb" to "vr".
+ * Once the front-end is updated to send "vr" as the homeLibraryCode, we can
+ * remove this mapping function.
+ */
+const mapEbToVr = (code) => {
+  if (code === "eb") {
```
russellcullen: nit: any chance we might receive `"EB"` or other non-lowercase variants? i.e. should we `code.toLowerCase()` for extra safety? Total nitpick because I feel like its not actually a problem in reality haha

russellcullen (approved):
LGTM! Feel free to ignore my paranoia 🤣

api/helpers/utils.js
Toxiapo: I think it's a valid point! Updating.

PR #200, opened 2026-04-20
Remove deprecated EB home branch

russellcullen:
## Description

We don't expect any more clients to send the old "EB" values -- we are now safe to remove that mapping.

## Checklist:


api/controllers/v0.3/endpoints.js
```diff
@@ -334,7 +330,7 @@ async function createPatron(req, res) {
     policy: Policy({ policyType }),
     ilsClient, // created above
     // "vr" is the default library code for digital cards
-    homeLibraryCode: mapEbToVr(req.body.homeLibraryCode) || "vr",
+    homeLibraryCode: req.body.homeLibraryCode || "vr",
```
Toxiapo, line 333: Thinking out loud here: Since `library card` is now a required field, I wonder if we should just remove this default and return an error to the client if this code is missing..For another ticket...

Toxiapo (approved):

api/controllers/v0.3/endpoints.js
russellcullen: You know I had the same thought 🤔 -- definitely worth considering and we can check on the requirement. I *think* ILS/Sierra is the part that "requires" (prefers?) a non-empty home library, so we should be safe to error here. And considering LCA doesn't allow it to be empty on the client side, I dont think there would be any functional downstream changes

PR #208, opened 2026-05-21
DO NOT MERGE: Remove default library (AI-generated)

entcheva:
## Description

Copilot CLI generated PR to remove default library attribute on the back end. This is for review & discussion purposes only, not ready to merge.

## Checklist:


api/models/v0.3/modelCard.js
```diff
@@ -91,6 +91,10 @@ class Card {
     // Now that we have all values, validate each field and if there are any
     // errors, collect them in the `errors` object which will be returned
     // to the user.
+    if (!this.homeLibraryCode || !this.homeLibraryCode.trim()) {
+      this.errors["homeLibraryCode"] = "A home library code is required.";
+    }
+
```
Toxiapo, line 97: Interesting! I was thinking to just "throw" the error as part of the "required field" check in line 70, but I think copilot is parsing the ticket literally to return an error here, which is also correct in the context of ticket description

